PR #18, opened 2022-04-08
Support for account alias

### LINKED ISSUE #15: Account aliases
**Task description**

Add ability to check whether addresses are aliases, and to generate aliases to the SDK.

EmilLa1:
## Purpose

Aims to fix #15 

## Changes

Expanded the AccountAddress type with the following two methods: 

`AccountAddress newAlias(int alias)`
Creates a new alias from an `AccountAddress`

`boolean isAliasOf(AccountAddress other)`
Checks whether an `AccountAddress` is an alias of the other.

There is now also a CHANGELOG.md in the root of the repository which should be used for the future.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-sdk/src/main/java/com/concordium/sdk/transactions/AccountAddress.java
+++ b/concordium-sdk/src/main/java/com/concordium/sdk/transactions/AccountAddress.java
@@ -1,16 +1,22 @@
 package com.concordium.sdk.transactions;
 
+import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.ToString;
 import lombok.val;
 
+import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
 import java.util.Objects;
 
 @ToString
+@EqualsAndHashCode
 public final class AccountAddress {
     public static final int BYTES = 32;
     private final static int VERSION = 1;
+    private final static int ACCOUNT_ADDRESS_PREFIX_SIZE = 29;
+    private final static int ALIAS_MAX_VALUE = 2 << 24;
 
     @Getter
     private final byte[] bytes;
@@ -30,6 +36,25 @@ public final class AccountAddress {
     public static AccountAddress from(String address) {
         val addressBytes = Base58.decodeChecked(VERSION, address);
         return AccountAddress.from(addressBytes);
+    }
+
+    public AccountAddress newAlias(int alias) {
+        if (alias < 0) {
+            throw new NumberFormatException("Alias must be non negative");
+        }
+        if (alias > ALIAS_MAX_VALUE) {
+            alias = alias % ALIAS_MAX_VALUE;
+        }
+        val buffer = ByteBuffer.allocate(Integer.BYTES);
+        buffer.putInt(alias);
+        val newAlias = this.bytes.clone();
+        System.arraycopy(buffer.array(), 1, newAlias, ACCOUNT_ADDRESS_PREFIX_SIZE, 3);
+        return AccountAddress.from(newAlias);
+    }
+
+    public boolean isAliasOf(AccountAddress other) {
+        return Arrays.equals(Arrays.copyOfRange(this.bytes, 0, ACCOUNT_ADDRESS_PREFIX_SIZE),
+                Arrays.copyOfRange(other.bytes, 0, ACCOUNT_ADDRESS_PREFIX_SIZE));
     }
 
     public static AccountAddress from(byte[] addressBytes) {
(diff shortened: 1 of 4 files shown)


concordium-sdk/src/main/java/com/concordium/sdk/transactions/AccountAddress.java
```diff
@@ -1,16 +1,22 @@
...
+import java.util.Arrays;
 import java.util.Objects;
 
 @ToString
+@EqualsAndHashCode
 public final class AccountAddress {
     public static final int BYTES = 32;
     private final static int VERSION = 1;
+    private final static int ACCOUNT_ADDRESS_PREFIX_SIZE = 29;
+    private final static int ALIAS_MAX_VALUE = 2 << 24;
```
abizjak, line 19: I don't think this is correct. This is `2^(1 + 24)`.

The max alias is `2^24 - 1`.

concordium-sdk/src/main/java/com/concordium/sdk/transactions/AccountAddress.java
```diff
@@ -32,6 +38,25 @@ public static AccountAddress from(String address) {
         return AccountAddress.from(addressBytes);
     }
 
+    public AccountAddress newAlias(int alias) {
+        if (alias < 0) {
+            throw new NumberFormatException("Alias must be non negative");
+        }
+        if (alias > ALIAS_MAX_VALUE) {
```
abizjak, line 45: If you set `ALIAS_MAX_VALUE = 1 << 24` then use `>= here`.

concordium-sdk/src/main/java/com/concordium/sdk/transactions/AccountAddress.java
```diff
@@ -32,6 +38,25 @@ public static AccountAddress from(String address) {
...
+            alias = alias % ALIAS_MAX_VALUE;
```
abizjak, line 46: This is a bad idea. I think you should raise an exception since otherwise they will silently get duplicate aliases.

concordium-sdk/src/main/java/com/concordium/sdk/transactions/AccountAddress.java
```diff
@@ -32,6 +38,25 @@ public static AccountAddress from(String address) {
...
+        }
+        val buffer = ByteBuffer.allocate(Integer.BYTES);
+        buffer.putInt(alias);
```
abizjak, line 49: Is this big-endian?

abizjak: It should ideally be big endian, since that is what we do in other SDKs.

concordium-sdk/src/main/java/com/concordium/sdk/transactions/AccountAddress.java
```diff
@@ -32,6 +38,25 @@ public static AccountAddress from(String address) {
...
+        val newAlias = this.bytes.clone();
+        System.arraycopy(buffer.array(), 1, newAlias, ACCOUNT_ADDRESS_PREFIX_SIZE, 3);
+        return AccountAddress.from(newAlias);
+    }
+
+    public boolean isAliasOf(AccountAddress other) {
+        return Arrays.equals(Arrays.copyOfRange(this.bytes, 0, ACCOUNT_ADDRESS_PREFIX_SIZE),
```
abizjak, line 56: Interesting, you have to copy arrays to compare parts of them? No slicing?

concordium-sdk/src/main/java/com/concordium/sdk/transactions/AccountAddress.java
EmilLa1: Yes big endian. https://docs.oracle.com/javase/7/docs/api/java/nio/ByteBuffer.html

concordium-sdk/src/main/java/com/concordium/sdk/transactions/AccountAddress.java
EmilLa1: I've set the `ALIAS_MAX_VALUE` to `(1 << 24 ) - 1`

concordium-sdk/src/main/java/com/concordium/sdk/transactions/AccountAddress.java
EmilLa1: I agree. I'll throw an exception to notify the user it has inputted a too large 'alias'.

concordium-sdk/src/main/java/com/concordium/sdk/transactions/AccountAddress.java
EmilLa1: I don't think there is an *easy* way with this java version at least.?

abizjak (approved):

PR #45, opened 2022-04-28
Support for configuring TLS.

EmilLa1:
## Purpose

Support for configuring TLS, mTLS etc.

## Changes

Introduced a `TLSConfig` type which can be used together with `Connection` for configuring TLS. 

Variants supported:

- 'auto' - will use the default jvm trust store. 
- 'custom' - TLS where the servers certificate or a CA authorizing it can be provided. 
- 'mTLS' - mutual TLS where the client is also provided with a certificate and a private key.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 
 ## Unreleased changes
+- Support for configuring TLS.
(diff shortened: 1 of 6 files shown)


abizjak (approved):
I'm surprised the TLSConfig is not some standard thing.

PR #49, opened 2022-04-29
Support for creating account address from regid

EmilLa1:
## Purpose

Support for creating an `AccountAddress` from the credential registration id. 

## Changes

- new type `CredentialRegistrationId`
- AccountAddress.from(CredentialRegistrationID) added

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-sdk/src/main/java/com/concordium/sdk/transactions/AccountAddress.java
+++ b/concordium-sdk/src/main/java/com/concordium/sdk/transactions/AccountAddress.java
@@ -49,6 +49,7 @@ public final class AccountAddress {
      * Create a {@link AccountAddress} from a {@link CredentialRegistrationId}.
      * @param regId the credential registration id to derive the {@link AccountAddress} from
      * @return the derived {@link AccountAddress}
+     * Note that this is only valid if the account was created via a credential with the given credential registration ID. That is, the derived address will only be correct for the account if the credential is the first credential on the account.
      */
     public static AccountAddress from(CredentialRegistrationId regId) {
         return AccountAddress.from(SHA256.hash(regId.getRegId()));
--- a/concordium-sdk/src/main/java/com/concordium/sdk/transactions/AccountAddress.java
+++ b/concordium-sdk/src/main/java/com/concordium/sdk/transactions/AccountAddress.java
@@ -1,5 +1,6 @@
 package com.concordium.sdk.transactions;
 
+import com.concordium.sdk.crypto.SHA256;
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
 import lombok.ToString;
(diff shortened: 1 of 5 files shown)


concordium-sdk/src/main/java/com/concordium/sdk/transactions/CredentialRegistrationId.java
```diff
@@ -0,0 +1,49 @@
...
+    private CredentialRegistrationId(byte[] regId) {
+        this.regId = regId;
+    }
+
+    /**
+     * Create a {@link CredentialRegistrationId} from the hex encoded registration id.
+     */
+    public static CredentialRegistrationId from(String encoded) {
+        try {
+            val regId = Hex.decodeHex(encoded);
```
abizjak, line 29: This is an API question, does this fail in cases like
```
aabb$
```
i.e., where the start is correct, but there is some leftover?

concordium-sdk/src/main/java/com/concordium/sdk/transactions/CredentialRegistrationId.java
```diff
@@ -0,0 +1,49 @@
...
+    }
+
+    /**
+     * Create a {@link CredentialRegistrationId} from raw bytes
+     *
+     * @param bytes the credential registration id bytes
+     * @return a {@link CredentialRegistrationId}
+     */
+    public static CredentialRegistrationId fromBytes(byte[] bytes) {
+        if (bytes.length != LENGTH) {
```
abizjak, line 43: In principle a credential registration ID is a group element in this G1 curve of the BLS pairing.

So not every series of bytes is a valid credential registration ID (in fact I think the probability that a random 48 bytes is a valid group element is somewhere around (1/2^17). I think this is fine for this version, but perhaps some documentation or "note for maintainers" could be added since in the future we might want to check validity.

concordium-sdk/src/main/java/com/concordium/sdk/transactions/AccountAddress.java
```diff
@@ -44,6 +45,15 @@ public static AccountAddress from(String address) {
         return AccountAddress.from(addressBytes);
     }
 
+    /**
+     * Create a {@link AccountAddress} from a {@link CredentialRegistrationId}.
+     * @param regId the credential registration id to derive the {@link AccountAddress} from
+     * @return the derived {@link AccountAddress}
```
abizjak, line 51:
```suggestion
     * Create a {@link AccountAddress} from a {@link CredentialRegistrationId}.
     * @param regId the credential registration id to derive the {@link AccountAddress} from
     * @return the derived {@link AccountAddress}
     * Note that this is only valid if the account was created via a credential with the given credential registration ID. That is, the derived address will only be correct for the account if the credential is the first credential on the account.
```

abizjak (approved):

concordium-sdk/src/main/java/com/concordium/sdk/transactions/CredentialRegistrationId.java
EmilLa1: Yes that is correct. I will add some documentation covering this subject. I don't think we need the implementation for verifying whether it's a group element of the group in question right now. 

So yes I agree :)

concordium-sdk/src/main/java/com/concordium/sdk/transactions/CredentialRegistrationId.java
EmilLa1: Yes. The Hex.decodeHex(hex) will throw a `DecoderException("Odd number of characters.")` which will be propagated via the `IllegalArgumentException` being thrown below.

PR #82, opened 2022-08-18
Query : Get Ancestors

parv0888:
## Purpose
Adds Support for Get Ancestors

## Changes
Adds Support for Get Ancestors

## Checklist

      hard-to-understand areas.

## CLA acceptance


concordium-sdk/src/main/java/com/concordium/sdk/Client.java
```diff
@@ -325,6 +326,29 @@ public Semver getVersion() {
...
     }
 
+    /**
+     * Gets Block Ancestor Blocks.
+     * @param blockHash {@link Hash} of the block.
+     * @param num Total no of Ancestor blocks to get.
+     * @return {@link BlockAncestors}
+     * @throws BlockNotFoundException When the returned response from Node is invalid or null.
+     */
+    public BlockAncestors getAncestors(Hash blockHash, long num) throws BlockNotFoundException {
```
EmilLa1, line 336: I think this should just return `ImmutableList<Hash>`.

concordium-sdk/src/main/java/com/concordium/sdk/responses/getancestors/BlockAncestors.java
```diff
@@ -0,0 +1,43 @@
...
+
+import java.util.Arrays;
+import java.util.stream.Collectors;
+
+@Builder
+@Getter
+/**
+ * Represents Ancestors of a Block.
+ */
+public class BlockAncestors {
```
EmilLa1, line 20: I don't think this abstraction provides much functionality, so I propose to simply delete it and return `ImmutableList<Hash>`.

README.md
```diff
@@ -220,6 +220,12 @@ Retrives the node software version.
 NodeInfo getNodeInfo()
 ```
 
+- `getAncestors`
+```java
+BlockAncestors getAncestors(Hash blockHash, long num) throws BlockNotFoundException
+```
+Gets Ancestor Blocks for the provided input Block Hash.
```
EmilLa1, line 227:
```suggestion
Get a list of block hashes that preceding the provided block hash and with a maximum size of the provided number.
```

EmilLa1 (changes requested):
Looks generally good, I don't think the `BlockAncestors` type brings much value to the table right now, so I suggest that we simply return `ImmutableList<Hash>`.

EmilLa1 (changes requested):
Even it's a simple query, I urge you to add a unit test for parsing the returned JSON.
To do that you probably want to factor out the deserialization logic into a separate helper function. 

Otherwise looks good.

parv0888:
> Even it's a simple query, I urge you to add a unit test for parsing the returned JSON. To do that you probably want to factor out the deserialization logic into a separate helper function.
> 
> Otherwise looks good.

Thanks, Added unit tests

EmilLa1 (approved):
Looks good

PR #140, opened 2022-10-02
Account Transaction : Transfer to Public

parv0888:
## Note
This transaction is part of https://github.com/Concordium/concordium-java-sdk/pull/149

## Purpose
Adds support for Account Transaction : `Transfer to Public`

## Changes
Adds support for Account Transaction : `Transfer to Public`

## Checklist

      hard-to-understand areas.

## CLA acceptance


crypto-jni/src/lib.rs
```diff
@@ -0,0 +1,288 @@
...
+        &mut csprng,
+    );
+
+    match payload {
+        Some(payload) => EncryptedTranfersResult::Ok(payload).to_jstring(&env),
+        None => Result::Err(PAYLOAD_CREATION_ERROR).to_jstring(&env),
+    }
+}
+
+static TABLE_BYTES: &[u8] = include_bytes!("table_bytes.bin");
```
EmilLa1, line 275: I'm not very keen on having this checked in like this. I don't think this should just magically be present. 

At least the babystepgiantstep table must be documented on how it was constructed.

EmilLa1 (changes requested):
Looks overall fine. 
Documentation is lacking and I think that 
all the encrypted transfer functionality should exist in the same PR as a user can't really do much  interesting without the full feature. 

Also I think the babystepgiantstep table should at least have some comprehensive documentation of how it is created and obtained. 

I do not think it should be checked in, but there should be something for creating it which is triggered by the makefile.

PR #158, opened 2023-01-26
GRPC V2 : GetBlocks & GetFinalizedBlocks

parv0888:
## Purpose
- Added support for GRPC V2 GetBlocks
- Added support for GRPC V2 GetFinalizedBlocks

## Changes
- Added support for GRPC V2 GetBlocks
- Added support for GRPC V2 GetFinalizedBlocks

## Checklist

      hard-to-understand areas.

## CLA acceptance


concordium-sdk/src/main/java/com/concordium/sdk/ClientV2.java
```diff
@@ -62,6 +64,61 @@ public Iterator<AnonymityRevokerInfo> getAnonymityRevokers(final BlockHashInput
...
+    /**
+     * Gets an {@link Iterator} of Finalized Blocks.
+     * With Specified Timeout.
+     * Form the time request is made and onwards.
+     * This can be used to listen for blocks being Finalized.
+     *
+     * @param timeoutMillis Timeout for the request in Milliseconds.
+     * @return {@link Iterator< BlockInfoV2 >}
+     */
+    public Iterator<BlockInfoV2> getFinalizedBlocks(int timeoutMillis) {
```
parv0888, line 115: Timeout has been included so that it can be overridden per request. 
That should be needed here because same client can be used to issue multiple requests and this method is specifically documented to be listening to incoming finalised blocks

concordium-sdk/src/main/java/com/concordium/sdk/ClientV2.java
```diff
@@ -62,6 +64,61 @@ public Iterator<AnonymityRevokerInfo> getAnonymityRevokers(final BlockHashInput
...
+    /**
+     * Gets an {@link Iterator} of Blocks Arriving at the node.
+     * With Specified Timeout.
+     * Form the time request is made and onwards.
+     * This can be used to listen for incoming blocks.
+     *
+     * @param timeoutMillis Timeout for the request in Milliseconds.
+     * @return {@link Iterator< BlockInfoV2 >}
+     */
+    public Iterator<BlockInfoV2> getBlocks(int timeoutMillis) {
```
parv0888, line 88: Timeout has been included so that it can be overridden per request.
That should be needed here because same client can be used to issue multiple requests and this method is specifically documented to be listening to incoming blocks

concordium-sdk/src/main/java/com/concordium/sdk/ClientV2.java
```diff
@@ -62,6 +64,61 @@ public Iterator<AnonymityRevokerInfo> getAnonymityRevokers(final BlockHashInput
         return to(grpcOutput, ClientV2MapperExtensions::to);
     }
 
+    /**
```
parv0888, line 67: To be removed

CHANGELOG.md
```diff
@@ -17,6 +16,9 @@
 - Added support for Configuring the account as a Delegator.
 - A new native dependency has been introduced in order to support the new transaction types mentioned above,
   As a result native dependencies need to be rebuilt before building the Java project.
+- Added initial support for GRPC V2
+- Added initial support for GRPC V2 query GetAnonymityRevokers
+- Added CLI example for GRPC V2 query GetAnonymityRevokers
```
EmilLa1, line 21: These changes should be put under "unreleased changes".

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/GetAnonymityRevokers.java
```diff
@@ -3,42 +3,43 @@
...
             names = {"--endpoint"},
             description = "GRPC interface of the node.",
             defaultValue = "http://localhost:20001")
     private String endpoint;
 
-
-    @SneakyThrows
     @Override
-    public void run() {
+    public Integer call() throws ClientInitializationException, MalformedURLException {
```
EmilLa1, line 25: Why not just `Runnable`?
Can it ever return some other value than `0`?

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/GetBlocks.java
```diff
@@ -0,0 +1,54 @@
...
+    private String endpoint;
+
+    @Option(
+            names = {"--timeout"},
+            description = "GRPC request timeout in milliseconds.",
+            defaultValue = "100000")
+    private int timeout;
+
+    @Override
+    public Integer call() throws MalformedURLException, ClientInitializationException {
```
EmilLa1, line 33: Same comment as above.

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/GetFinalizedBlocks.java
```diff
@@ -0,0 +1,52 @@
...
+    private String endpoint;
+
+    @Option(
+            names = {"--timeout"},
+            description = "GRPC request timeout in milliseconds.",
+            defaultValue = "100000")
+    private int timeout;
+
+    @Override
+    public Integer call() throws MalformedURLException, ClientInitializationException {
```
EmilLa1, line 31: Same as above.

concordium-sdk/src/main/java/com/concordium/sdk/responses/BlockInfoV2.java
```diff
@@ -0,0 +1,29 @@
...
+import lombok.ToString;
+
+/**
+ * Information about the Block. See {@link ClientV2#getBlocks()} & {@link ClientV2#getFinalizedBlocks()} ()}.
+ */
+@RequiredArgsConstructor
+@Builder
+@ToString
+@EqualsAndHashCode
+public class BlockInfoV2 {
```
EmilLa1, line 18: Why can't the existing `BlockInfo` be used?

EmilLa1: Also there seem to be missing some data?
https://github.com/Concordium/concordium-java-sdk/blob/main/concordium-sdk/src/main/java/com/concordium/sdk/responses/blockinfo/BlockInfo.java

concordium-sdk/src/main/java/com/concordium/sdk/responses/blocksummary/updates/queues/AnonymityRevokerInfo.java
```diff
@@ -28,6 +28,7 @@ public final class AnonymityRevokerInfo {
      */
     private final Description description;
 
+    @ToString.Exclude
```
EmilLa1, line 31: Why exclude this from the `toString()`?

EmilLa1 (changes requested):

concordium-sdk/src/main/java/com/concordium/sdk/responses/blocksummary/updates/queues/AnonymityRevokerInfo.java
parv0888: Public Key was large while trying to print the Object in Console. Hence removed it just to beautify the tostring method output. 
The assumption here is that Public Key as an Object is not useful when printed on a terminal.

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/GetAnonymityRevokers.java
parv0888: Call here is not being used to specify a return code. Returning an exit code is a side effect which might be useful in further development. 

Currently its being used so that Checked Exceptions can be added to the Method signature. Since earlier approach to use `@SneakyThrows` in the Examples code was not an Elegant Solution. 
The Library used to create console application automatically handles the Exceptions and prints them to the console.

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/GetBlocks.java
parv0888: Answered above. 

"Call here is not being used to specify a return code. Returning an exit code is a side effect which might be useful in further development.

Currently its being used so that Checked Exceptions can be added to the Method signature. Since earlier approach to use @SneakyThrows in the Examples code was not an Elegant Solution.
The Library used to create console application automatically handles the Exceptions and prints them to the console."

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/GetFinalizedBlocks.java
parv0888: Answered Above

"Call here is not being used to specify a return code. Returning an exit code is a side effect which might be useful in further development.

Currently its being used so that Checked Exceptions can be added to the Method signature. Since earlier approach to use @SneakyThrows in the Examples code was not an Elegant Solution.
The Library used to create console application automatically handles the Exceptions and prints them to the console."

CHANGELOG.md
parv0888: fixed

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/GetAnonymityRevokers.java
parv0888: @EmilLa1  - please do suggest if we need to revert back to the previous approach.

EmilLa1: Well, I don't see why one would care about checked exceptions in this particular cli example. And as it is just an example, then I would prefer it to be as simple as possible. 

But I'll let you decide.

concordium-sdk/src/main/java/com/concordium/sdk/responses/blocksummary/updates/queues/AnonymityRevokerInfo.java
EmilLa1: Isn't it just a hex string? 

I mean, the purpose of the `toString` is mainly debugging - and here it is usually good practice to include all the fields.

concordium-sdk/src/main/java/com/concordium/sdk/responses/BlockInfoV2.java
EmilLa1, line 18: I suggest naming this otherwise as not to confuse it with `BlockInfo` as I just did. 

How about creating a class `BlockIdentifier` which holds these information, and then use that also in the existing `BlockInfo` ? (    private Integer blockHeight and private Hash blockHash; )

And note that the `BlockInfo` should still expose getBlockHeight() and getBlockHash() - which would then just call into the `BlockIdentifier`.

And then it should also be possible to pass a `BlockIdentifer` to the grpc in order to request a block. Maybe the `BlockIdentifier` should have a function which would return the response type you're using there.

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/GetAnonymityRevokers.java
parv0888: Thanks @EmilLa1

The Only option barring using `@SneakyThrows` is to use a `Callable` instead of a `Runnable` and handle the errors.
We should be able to use the return code also in later examples as a way to provide return code when needed for a Runtime exception. 
Allow us to keep it as it is because it provides better and more read-able code structure.

concordium-sdk/src/main/java/com/concordium/sdk/responses/BlockInfoV2.java
parv0888: BlockInfoV2 has been renamed to BlockIdentifier.
BlockInfo is not derived from BlockIdentifier.

Regarding using BlockIdentifier as an input to fetch blocks. 
At various places where block and related information is being requested the API input parameter is a `Hash` (block hash). 
We can create an Overloaded method to also include `BlockIdentifier` since `BlockIdentifier` also has `blockHeight` which may or may not be available with Block`Hash` but that might not be necessary as Hash can be retrived from `BlockIdentifier`.

concordium-sdk/src/main/java/com/concordium/sdk/responses/blocksummary/updates/queues/AnonymityRevokerInfo.java
parv0888: fixed

concordium-sdk/src/main/java/com/concordium/sdk/responses/BlockIdentifier.java
```diff
@@ -0,0 +1,31 @@
...
+import com.concordium.sdk.transactions.Hash;
+import com.concordium.sdk.types.UInt64;
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import lombok.ToString;
+import lombok.experimental.SuperBuilder;
+
+/**
+ * Information about the Block. See {@link ClientV2#getBlocks()} & {@link ClientV2#getFinalizedBlocks()} ()}.
```
EmilLa1, line 13:
```suggestion
 * Identification of a block. See {@link ClientV2#getBlocks()} & {@link ClientV2#getFinalizedBlocks()} ()}.
```

concordium-sdk/src/main/java/com/concordium/sdk/responses/blockinfo/BlockInfo.java
```diff
@@ -1,90 +1,89 @@
...
-@RequiredArgsConstructor
 @Getter
-public class BlockInfo {
-    /**
-     *  Hash of the block.
-     */
-    private Hash blockHash;
+@SuperBuilder
+@Jacksonized
+public class BlockInfo extends BlockIdentifier {
```
EmilLa1, line 18: I think it should simply be a field on the `BlockInfo`, and then you manually create the two `getBlockHash` and `getBlockHeight` functions such that the API becomes backwards compatible. 

Whether these two fields are kept as two fields on inside the `BlockIdentifier` should be an implementation detail.

concordium-sdk/src/main/java/com/concordium/sdk/responses/BlockInfoV2.java
EmilLa1: Please see this comment. https://github.com/Concordium/concordium-java-sdk/pull/158#discussion_r1108183961 We should remain backwards compatible. 


Regarding the block identifier and using for querying calls. Let's just keep as is for now.

concordium-sdk/src/main/java/com/concordium/sdk/responses/BlockIdentifier.java
parv0888: done

concordium-sdk/src/main/java/com/concordium/sdk/responses/blockinfo/BlockInfo.java
parv0888: This has been fixed by making `BlockIdentifier` a field in `BlockInfo`

EmilLa1 (approved):
Looks good.

PR #160, opened 2023-02-16
GRPC V2 : Get Account Info

parv0888:
## Purpose
Added support for GRPC V2 `getAccountInfo`

## Changes
Added support for GRPC V2 `getAccountInfo`

## Checklist

      hard-to-understand areas.

## CLA acceptance


concordium-sdk/src/main/java/com/concordium/sdk/ClientV2MapperExtensions.java
```diff
@@ -100,4 +128,332 @@ static BlockIdentifier to(FinalizedBlockInfo finalizedBlockInfo) {
...
+        return AccountIndex.newBuilder().setValue(index.getIndex().getValue()).build();
+    }
+
+    static AccountAddress to(com.concordium.sdk.transactions.AccountAddress address) {
+        return AccountAddress.newBuilder()
+                .setValue(to(address.getBytes()))
+                .build();
+    }
+
+    static com.concordium.sdk.responses.accountinfo.AccountInfo to(AccountInfo account) {
```
EmilLa1, line 166: I think this function should be a static member of `AccountInfo` and then called `from`.

Otherwise I think this file will get out of hand.

EmilLa1: And the same comment for similar functions present here.  Please move them out to the respective classes and make `from` functions there. I think that will be easier to manage in the end.

concordium-sdk/src/main/java/com/concordium/sdk/ClientV2MapperExtensions.java
```diff
@@ -100,4 +128,332 @@ static BlockIdentifier to(FinalizedBlockInfo finalizedBlockInfo) {
...
+                .accountEncryptionKey(ElgamalPublicKey.from(account.getEncryptionKey().getValue().toByteArray()))
+                .accountIndex(to(account.getIndex()))
+                .accountAddress(to(account.getAddress()));
+
+        if (account.hasStake()) {
+            if (account.getStake().hasBaker()) {
+                builder.bakerInfo(to(account.getStake().getBaker()));
+            } else if (account.getStake().hasDelegator()) {
+                builder.delegation(to(account.getStake().getDelegator()));
+            } else {
```
EmilLa1, line 186: You don't need need case I think. At least it should have a proper error message. 

But if you look at the `.proto` definition then it's a `oneof` 
```
oneof staking_info {
    // The account is a baker.
    Baker baker = 1;
    // The account is a delegator.
    Delegator delegator = 2;
  }
```

So I think a simple `if..else` is appropriate, because if the account is staking then it's either a baker or a delegator.

concordium-sdk/src/main/java/com/concordium/sdk/crypto/ed25519/ED25519PublicKey.java
```diff
@@ -2,36 +2,42 @@
...
         this.bytes = bytes;
     }
 
+    public byte[] getBytes() {
+        return copyOf(bytes, bytes.length);
+    }
+
     @JsonCreator
-    public static ED25519PublicKey from(String hexKey) {
+    public static ED25519PublicKey from(String hexKey) throws ED25519Exception {
```
EmilLa1, line 27: There is no reason to declare the `ED25519Exception` here as it is **now** unchecked. (See my comment below)
Further it is not thrown by this function, but from the one below. 

I suggest removing the `throws` from this function, 
and declare it in the java doc of the function below. 

And also please add documentation to these public functions. What do they expect what to do they return, when do they throw `IllegalArgumentException` and `ED25519Exception` .

concordium-sdk/src/main/java/com/concordium/sdk/exceptions/ED25519Exception.java
```diff
@@ -3,7 +3,7 @@
 import com.concordium.sdk.crypto.ed25519.ED25519ResultCode;
 import lombok.Getter;
 
-public final class ED25519Exception extends Exception {
+public final class ED25519Exception extends RuntimeException {
```
EmilLa1, line 6: What is the reasoning of making this an unchecked exception instead of the prior checked one?

The thing is, when you change this to be an unchecked one then you must document the functions that throws it with java doc. But when it is checked it is also clear from the type what the contract of the function is. 

I am as such not opposed to using an unchecked exception here as there is little the user can do. But please document all functions now throwing this with proper java doc.

concordium-sdk/src/main/java/com/concordium/sdk/responses/accountinfo/AccountEncryptedAmount.java
```diff
@@ -16,13 +21,45 @@
...
+     * If present, the amount that has resulted from aggregating other amounts
+     * If this field is present so is `num_aggregated`.
+     */
+    private final Optional<EncryptedAmount> aggregatedAmount;
+
+    /**
+     * The number of aggregated amounts (must be at least 2 if present). This
+     * field is present if and only if `aggregated_amount` is present.
+     */
+    private final Optional<Integer> numAggregated;
```
EmilLa1, line 64: Great job here! Super useful documentation for the user. :+1:

concordium-sdk/src/main/java/com/concordium/sdk/responses/accountinfo/credential/AttributeType.java
```diff
@@ -0,0 +1,34 @@
+package com.concordium.sdk.responses.accountinfo.credential;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public enum AttributeType {
```
EmilLa1, line 5: Please add a short doc at the enum declaration about what the `AttributeType` is reflecting.

concordium-sdk/src/main/java/com/concordium/sdk/responses/nodeinfo/NodeInfo.java
```diff
@@ -19,7 +19,7 @@
 @ToString
 public class NodeInfo {
 
-    private static final ZoneId UTC_ZONE = ZoneId.of("UTC");
+    public static final ZoneId UTC_ZONE = ZoneId.of("UTC");
```
EmilLa1, line 22: This should probably be moved into a separate file now it's used more places.

concordium-sdk/src/test/java/com/concordium/sdk/ClientV2GetAccountInfoTest.java
```diff
@@ -0,0 +1,474 @@
...
+                            ByteString.copyFrom(FIRST_NAME_VALUE)))
+                    .build());
+
+    private static Commitment toCommitment(byte[] commitmentCredCounter) {
+        return Commitment.newBuilder()
+                .setValue(ByteString.copyFrom(commitmentCredCounter))
+                .build();
+    }
+
+    private static final Credential NORMAL_CREDENTIAL_CLIENT =
```
EmilLa1, line 140:
```suggestion
    private static final Credential NORMAL_CREDENTIAL_EXPECTED =
```

Because this is the expectation right?

concordium-sdk/src/test/java/com/concordium/sdk/ClientV2GetAccountInfoTest.java
```diff
@@ -0,0 +1,474 @@
...
+    private static final byte[] COMMITMENT_ID_CRED_SHARING_COEFF = new byte[]{12, 12, 12};
+    private static final byte[] COMMITMENT_ATTRIBUTE_0 = new byte[]{13, 13, 13};
+    private static final YearMonth POLICY_CREATED_AT
+            = YearMonth.newBuilder().setYear(2022).setMonth(10).build();
+    private static final YearMonth POLICY_VALID_TO
+            = YearMonth.newBuilder().setYear(2024).setMonth(10).build();
+    private static final byte[] FIRST_NAME_VALUE
+            = "policy-value-0".getBytes(StandardCharsets.UTF_8);
+    private static final byte[] ENC_ID_PUB_SHARE = new byte[]{90, 90, 90};
+    private static final NormalCredentialValues.Builder NORMAL_CREDENTIAL_BUILDER = NormalCredentialValues.newBuilder()
```
EmilLa1, line 102: Why return the builder? You never extend it?

EmilLa1: This is the mocked return value? Which you are checking against the "expected" result declared below?

EmilLa1: Maybe name it NORMAL_CREDENTIAL_RESULT or something similar.

concordium-sdk/src/test/java/com/concordium/sdk/ClientV2GetAccountInfoTest.java
```diff
@@ -0,0 +1,474 @@
...
+import static com.concordium.sdk.responses.accountinfo.credential.AttributeType.FIRST_NAME;
+import static com.concordium.sdk.responses.nodeinfo.NodeInfo.UTC_ZONE;
+import static com.concordium.sdk.transactions.CredentialRegistrationId.fromBytes;
+import static java.time.YearMonth.of;
+import static org.junit.Assert.assertEquals;
+import static org.mockito.AdditionalAnswers.delegatesTo;
+import static org.mockito.Mockito.*;
+
+@RunWith(MockitoJUnitRunner.class)
+public class ClientV2GetAccountInfoTest {
```
EmilLa1, line 53: Great testing. :+1:

EmilLa1 (commented):
Looks good. :+1:  A few minor comments here and there.

concordium-sdk/src/main/java/com/concordium/sdk/ClientV2MapperExtensions.java
parv0888: @EmilLa1  - Yes the file is getting bigger. 
Can i suggest breaking this file into smaller files with non static functions. Then putting them in the ClientV2 Constructor. 
Wanted to keep the serialization / deserialization logic out of the respective classes. Simillar to what Object Mapper and other. Basically meaning that the Client Class dosent know of any Proto class.

concordium-sdk/src/main/java/com/concordium/sdk/exceptions/ED25519Exception.java
parv0888: done

concordium-sdk/src/main/java/com/concordium/sdk/crypto/ed25519/ED25519PublicKey.java
parv0888: done

concordium-sdk/src/main/java/com/concordium/sdk/responses/nodeinfo/NodeInfo.java
parv0888: moved to a separate constants file.

concordium-sdk/src/test/java/com/concordium/sdk/ClientV2GetAccountInfoTest.java
parv0888: removed it and merged with the Account Info Creation since there was only a single usage of the value.

concordium-sdk/src/test/java/com/concordium/sdk/ClientV2GetAccountInfoTest.java
parv0888: thanks

concordium-sdk/src/test/java/com/concordium/sdk/ClientV2GetAccountInfoTest.java
parv0888: removed and merge with the main Account Info Building / Instantiation. Since there was only a single usage

concordium-sdk/src/main/java/com/concordium/sdk/ClientV2MapperExtensions.java
parv0888: Can we create an issue for this and then fix it. 
Basically we want to create a non static class. And Client then depends on an object of this file.
This would allow for better testing since in bigger Objects link AccountInfo there are multiple branches which should be tested and its better that the mapping code is tested independently of the Client code.

EmilLa1: Yes. I think that is OK. :+1:

EmilLa1 (approved):
Looks good.

concordium-sdk/src/main/java/com/concordium/sdk/responses/accountinfo/credential/AttributeType.java
EmilLa1: Can you add a short doc here and then it's good to go. Thanks

EmilLa1 (approved):

PR #229, opened 2023-04-23
GRPC V2 : Get Branches

parv0888:
## Purpose

Added support for GRPC V2 : Get Branches

## Changes

Added support for GRPC V2 : Get Branches

## Checklist

      hard-to-understand areas.

## CLA acceptance


concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/GetBranches.java
```diff
@@ -0,0 +1,50 @@
...
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.concurrent.Callable;
+
+@Command(name = "GetBranches", mixinStandardHelpOptions = true)
+public class GetBranches implements Callable<Integer> {
+    @Option(
+            names = {"--endpoint"},
+            description = "GRPC interface of the node.",
+            defaultValue = "http://localhost:20002")
```
abizjak, line 25: That's an odd default value. Please use 20000

concordium-sdk/src/main/java/com/concordium/sdk/ClientV2MapperExtensions.java
```diff
@@ -1245,4 +1246,11 @@ static com.concordium.sdk.responses.blockinfo.BlockInfo to(BlockInfo blockInfo)
                 .blockArriveTime(to(to(blockInfo.getArriveTime())))
                 .build();
     }
+
+    static Branch to(com.concordium.grpc.v2.Branch branch) {
+        return Branch.builder()
+                .blockHash(to(branch.getBlockHash()))
+                .children(to(branch.getChildrenList(), ClientV2MapperExtensions::to))
+                .build();
```
abizjak, line 1254: How deeply can this be nested? This is a recursive call on data that you are getting from the node.

I assume at some point this will cause a stack overflow when there is too much branching. When will that happen?

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/GetBranches.java
parv0888: done

concordium-sdk/src/main/java/com/concordium/sdk/ClientV2MapperExtensions.java
parv0888: The actual stack size depends on the OS and can be manipulated.
Do you want us to change the return type to a Class backed by a graph / adjacency list?

abizjak: That would be ideal, but it would add a lot of complexity. 

Maybe it is sufficient for now to just comment that if there is a lot of branching this method can cause a stack overflow.

EmilLa1 (approved):

PR #270, opened 2023-09-21
Deserializing fix

EmilLa1:
## Purpose

Retaining the `Payload` when deserializing a transaction from bytes.

This makes deserializing safer as one can now guard via the `TransactionType` before casting the payload to the concrete type.

Also fixed a bug where the maximum energy allowed to spend on the transaction (for smart contract related transactions) was actually set higher than what the user set. In practice the cost of checking signatures was added to what the user had specified)

## Checklist

      hard-to-understand areas.


concordium-sdk/src/main/java/com/concordium/sdk/transactions/RawPayload.java
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * The bytes of the raw payload.
+     */
+
+    @Getter
+    private final byte[] rawBytes;
+
+    public RawPayload(final byte[] bytes) {
+        this.rawBytes = bytes;
+    }
```
abizjak, line 23: Shouldn't you be checking here that there is at least 1 byte for the tag? And ideally that it's a valid tag?

concordium-sdk/src/main/java/com/concordium/sdk/transactions/TransactionType.java
```diff
@@ -40,4 +40,13 @@ public enum TransactionType {
     TransactionType(byte type) {
         this.value = type;
     }
+
+    public static TransactionType parse(byte type) {
+        for (TransactionType transactionType : values()) {
+            if (transactionType.value == type) {
+                return transactionType;
+            }
+        }
```
abizjak, line 49: This makes me a bit sad :( 

Iterating through all the values to find the one we want when we in principle know the tags.

concordium-sdk/src/main/java/com/concordium/sdk/transactions/TransactionType.java
```diff
@@ -40,4 +40,13 @@ public enum TransactionType {
...
+        throw new IllegalArgumentException("Unknown transaction type tag");
```
abizjak, line 50: Probably useful to add the `byte` to this exception for easier debugging.

abizjak (approved):

concordium-sdk/src/main/java/com/concordium/sdk/transactions/RawPayload.java
EmilLa1: Good idea.

concordium-sdk/src/main/java/com/concordium/sdk/transactions/TransactionType.java
EmilLa1: Good idea.

concordium-sdk/src/main/java/com/concordium/sdk/transactions/TransactionType.java
EmilLa1: me too  :(

PR #274, opened 2023-10-11
Smart contract schema support

magnusbechwind:
## Purpose

Adds functionality for creating and serializing smart contract parameters as opposed to only allowing raw bytes as parameters.

## Changes

Extended the FFI bindings to allow for passing parameters as json to the rust library and getting the serialized parameter back.

Added `JNINError` class allowing for error messages to be returned from the rust library.

Created a class, `Schema`, to represent smart contract schemas.

Created a class `SchemaParameter` to represent parameters serializable via a `Schema`.

Added from methods to `InitContractPayload` and `UpdateContractPayload` accepting the new `SchemaParameter`.

Implemented custom serialization of `AbstractAddress` and added `ListParam`, `AccountAddressParam`, `AddressParam` and `ContractAddressParam` for conveniently using lists of objects, `AbstractAddress`, `ContractAddress` and `AccountAddress` as smart contract parameters.

Extended the examples with `Cis2Nft` and `Cis2WCCD` demonstrating how to instantiate parameters for the different smart contract methods present in the cis2-nft and cis2-wCCD contracts.

## Checklist

      hard-to-understand areas.


concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/cis2nft/Cis2NftParameters.java
```diff
@@ -0,0 +1,140 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Helper class for creating and initializing parameters being used in {@link Cis2Nft}.
+ */
+public class Cis2NftParameters {
```
EmilLa1, line 20: Maybe add some documentation that links to the smart contract source code. This will help a user "bridge" the rust parameters with the java parameters I think.

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/cis2nft/Cis2Nft.java
```diff
@@ -0,0 +1,162 @@
...
+                .sender(AccountAddress.from(SENDER_ADDRESS))
+                .payload(payload)
+                .expiry(EXPIRY)
+                .nonce(AccountNonce.from(nonce))
+                .signer(SIGNER)
+                .maxEnergyCost(UInt64.from(10000))
+                .build();
+        Hash txHash = client.sendTransaction(transaction);
+        System.out.println("Submitted transaction for " + this.methodName  + " with hash: " + txHash);
+        sleep();
```
EmilLa1, line 143: This `sleep` is a bit hacky. 

Instead what you could do, is to query the transaction status (and block) until it's finalized. The function should also take a timeout `Duration` such that we don't end up in a situation where it blocks forever.

This could actually be a neat little helper function you could add to the `ClientV2`. We have something similar for the rust sdk.

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/cis2nft/Cis2Nft.java
```diff
@@ -0,0 +1,162 @@
...
+    @SneakyThrows
+    private void sleep() {
+        for (int i = 0; i < 200; i++) {
+            System.out.print(".");
+            if (i % 50 == 0) {
+                System.out.println("\n");
+            }
+            Thread.sleep(30);
+        }
+    }
```
EmilLa1, line 157:
```suggestion
```

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/cis2nft/Cis2NftMethod.java
```diff
@@ -0,0 +1,16 @@
+package com.concordium.sdk.examples.contractexample.cis2nft;
+
+/**
+ * Represents the different possible methods able to be run by {@link Cis2Nft}
+ */
+public enum Cis2NftMethod {
```
EmilLa1, line 6: Link to the source code (and document what the different values mean)

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/cis2nft/Cis2NftParameters.java
```diff
@@ -0,0 +1,140 @@
...
+        NftTransfer transfer = new NftTransfer(tokenId, amount, from, to, data);
+        List<NftTransfer> transfers = new ArrayList<>();
+        transfers.add(transfer);
+        SchemaParameter transferParameter = new NftTransferParam(schema, nftTransferReceiveName, transfers);
+        transferParameter.initialize(true);
+
+        return transferParameter;
+    }
+
+    @SneakyThrows
```
EmilLa1, line 61: I would prefer being consistent here with the `@SneakyThrows`

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/cis2nft/Cis2NftParameters.java
```diff
@@ -0,0 +1,140 @@
...
+        List<UpdateOperator> updateOperatorList = new ArrayList<>();
+        updateOperatorList.add(update1);
+        updateOperatorList.add(update2);
+        SchemaParameter updateOperatorsParams = new UpdateOperatorParams(schema, updateOperatorReceiveName, updateOperatorList);
+        updateOperatorsParams.initialize(true);
+
+        return updateOperatorsParams;
+    }
+
+    public static SchemaParameter generateOperatorOfParams() throws IOException {
```
EmilLa1, line 76: See above comment.

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/cis2nft/Cis2NftParameters.java
```diff
@@ -0,0 +1,140 @@
...
+        ReceiveName setImplementorsReceiveName = ReceiveName.from(CIS_2_NFT_CONTRACT_NAME, "setImplementors");
+        List<ContractAddress> implementors = new ArrayList<>();
+        String identifier = "IdentifierID";
+        implementors.add(CONTRACT_ADDRESS_1);
+        implementors.add(CONTRACT_ADDRESS_2);
+        SchemaParameter setImplementorsParams = new SetImplementorsParams(schema, setImplementorsReceiveName, identifier, implementors);
+        setImplementorsParams.initialize();
+        return setImplementorsParams;
+    }
+
```
EmilLa1, line 139: also see above comment.

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/cis2nft/Cis2NftParameters.java
```diff
@@ -0,0 +1,140 @@
...
+        tokens.add(TokenIdU32.from(2));
+        tokens.add(TokenIdU32.from(22));
+        tokens.add(TokenIdU32.from(2132));
+        SchemaParameter mintParameter = new MintParams(schema, mintParamsReceiveName, ACCOUNT_ADDRESS, tokens);
+        mintParameter.initialize(true);
+        return mintParameter;
+    }
+
+    @SneakyThrows
+    public static SchemaParameter generateTransferParams() {
```
EmilLa1, line 42: Please add a bit context to each of these functions.

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/parameters/BalanceOfQuery.java
```diff
@@ -0,0 +1,19 @@
...
+import lombok.Getter;
+
+/**
+ * Represents a '<a href = "https://github.com/Concordium/concordium-rust-smart-contracts/blob/main/concordium-cis2/src/lib.rs#L1138">BalanceOfQuery</a>' used in different smart contracts
+ */
+@Getter
+@AllArgsConstructor
+public class BalanceOfQuery<T extends TokenId> {
+    @JsonProperty("token_id")
+    private T tokenId;
```
EmilLa1, line 16: Document fields please. 

Imagine a developer sees this and he has 0 experience with the SDK. 

For example, "What is an `AbstractAddress`". Well, it's either an address of a contract or an account.

EmilLa1 (commented):
I am not all the way through yet - but so far looks good.

crypto-jni/src/lib.rs
```diff
@@ -179,28 +182,83 @@ pub extern "system" fn Java_com_concordium_sdk_crypto_CryptoJniNative_generatePu
 #[derive(SerdeSerialize, SerdeDeserialize)]
 enum CryptoJniResult<T> {
     Ok(T),
-    Err(jint),
+    Err(JNIError),
+}
+
+#[derive(SerdeSerialize, SerdeDeserialize)]
+#[allow(non_snake_case)]
+enum JNIErrorType {
```
EmilLa1, line 190: So this is more of a status code than an actual `Error` right? 

I suggest renaming it to something that differentiates it from "normal" rust errors.

crypto-jni/src/lib.rs
```diff
@@ -408,3 +467,199 @@ pub extern "system" fn Java_com_concordium_sdk_crypto_CryptoJniNative_generateCo
...
     return CryptoJniResult::Ok(payload).to_jstring(&env);
 }
+
+type SerializeParamResult = CryptoJniResult<String>;
+
+#[no_mangle]
+#[allow(non_snake_case)]
+/**
+ * The JNI wrapper for serializing receive parameters.
+ * Constructs a SerializeParamResult containg the hex encoded serialized parameter.
```
EmilLa1, line 477: what if it fails? Can you add a few more sentences to this. 

Also perhaps some preconditions of the inputted parameters?

crypto-jni/src/lib.rs
```diff
@@ -408,3 +467,199 @@ pub extern "system" fn Java_com_concordium_sdk_crypto_CryptoJniNative_generateCo
...
+        },
+        Err(err) => return SerializeParamResult::from(err).to_jstring(&env),
+    };
+
+    let version: Option<u8> = match schemaVersion {
+        0 => Some(0),
+        1 => Some(1),
+        2 => Some(2),
+        3 => Some(3),
+        _ => None,
```
EmilLa1, line 525: Is this sensible? What if the provided schema version was 4?

crypto-jni/src/lib.rs
```diff
@@ -408,3 +467,199 @@ pub extern "system" fn Java_com_concordium_sdk_crypto_CryptoJniNative_generateCo
...
+    );
+
+
+    let result = match serializedParameter {
+        Ok(serializedParameter) => hex::encode(serializedParameter),
+        Err(err) => return SerializeParamResult::from(err).to_jstring(&env),
+        
+    };
+
+    return SerializeParamResult::Ok(result).to_jstring(&env);
```
EmilLa1, line 546: So errors are also "Ok" ?

concordium-sdk/src/test/testresources/smartcontractschema/lib.rs
```diff
@@ -0,0 +1,242 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+//! # A Concordium V1 smart contract
```
EmilLa1, line 3: Please add some more documentation to this SC. 

For example, what is the purpose of it? Why is it here in the java sdk?

concordium-sdk/src/test/testresources/smartcontractschema/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+fn init<S: HasStateApi>(
+    _ctx: &impl HasInitContext,
+    _state_builder: &mut StateBuilder<S>,
+) -> InitResult<State> {
+    // Your code
+
+    Ok(State {})
+}
+
+//
```
EmilLa1, line 34: // ?

concordium-sdk/src/test/testresources/smartcontractschema/lib.rs
```diff
@@ -0,0 +1,242 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+//! # A Concordium V1 smart contract
+use concordium_std::*;
+use core::fmt::Debug;
+
+/// Your smart contract state.
+#[derive(Serialize, SchemaType)]
+pub struct State {
+    // Your state
```
EmilLa1, line 10: Which state?

concordium-sdk/src/test/testresources/smartcontractschema/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+}
+
+/// Your smart contract errors.
+#[derive(Debug, PartialEq, Eq, Reject, Serial, SchemaType)]
+enum Error {
+    /// Failed parsing the parameter.
+    #[from(ParseError)]
+    ParseParams,
+    /// Your error
+    YourError,
```
EmilLa1, line 20: What is "YourError" ?

concordium-sdk/src/test/testresources/smartcontractschema/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+
+/// Your smart contract state.
+#[derive(Serialize, SchemaType)]
+pub struct State {
+    // Your state
+}
+
+/// Your smart contract errors.
+#[derive(Debug, PartialEq, Eq, Reject, Serial, SchemaType)]
+enum Error {
```
EmilLa1, line 15: Maybe this should be called something a bit less generic. E.g. "TestContractError" ?

concordium-sdk/src/test/testresources/smartcontractschema/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+    Ok(())
+}
+
+/// Takes Address as parameter for ensuring correct serialization of Address when passed directly
+#[receive(
+    contract = "java_sdk_schema_unit_test",
+    name = "account_address_test",
+    parameter = "AccountAddress",
+    error = "Error",
+    mutable
```
EmilLa1, line 155: why mutable?

EmilLa1: (Question also goes for the other receive functions here)

EmilLa1 (commented):
A few more comments for now.

concordium-sdk/src/main/java/com/concordium/sdk/transactions/smartcontracts/parameters/ListParam.java
```diff
@@ -0,0 +1,47 @@
...
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import lombok.Getter;
+
+import java.io.IOException;
+import java.util.List;
+
+/**
+ * Wrapper for passing lists as smart contract parameters.<p>
```
EmilLa1, line 18: `<p>` ?

concordium-sdk/src/main/java/com/concordium/sdk/types/UInt8.java
```diff
@@ -0,0 +1,57 @@
...
+import lombok.EqualsAndHashCode;
+import lombok.Getter;
+import lombok.val;
+
+import java.io.IOException;
+
+@EqualsAndHashCode
+@Getter
+@JsonSerialize(using = UInt8.UInt8Serializer.class)
+public class UInt8 {
```
EmilLa1, line 16: ... an unsigned byte

concordium-sdk/src/test/java/com/concordium/sdk/smartcontract/IncorrectParams.java
```diff
@@ -0,0 +1,17 @@
+package com.concordium.sdk.smartcontract;
+
+import com.concordium.sdk.transactions.ReceiveName;
+import com.concordium.sdk.transactions.smartcontracts.Schema;
+import com.concordium.sdk.transactions.smartcontracts.SchemaParameter;
+import lombok.Getter;
+
+@Getter
+public class IncorrectParams extends SchemaParameter {
```
EmilLa1, line 9: documentation of this please.

concordium-sdk/src/main/java/com/concordium/sdk/transactions/smartcontracts/parameters/ListParam.java
EmilLa1: This should be used only when actual parameter is a list/vector right? Otherwise if the list is somewhere contained within the parameter, then it's fine to just use a plain old `List<Foo>` right?

EmilLa1 (commented):
Looks good! A few things here and there I believe we can make even better.

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/cis2nft/Cis2NftMethod.java
magnusbechwind: How in-depth should the explanation of the values be? Perhaps i should link the parameters each value is used with?

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/cis2nft/Cis2NftParameters.java
magnusbechwind: image

I have done something like this for every method. Does it need to be more detailed?

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/parameters/Transfer.java
```diff
@@ -0,0 +1,28 @@
...
+
+import java.util.List;
+
+
+/**
+ * Represents a '<a href = "https://github.com/Concordium/concordium-rust-smart-contracts/blob/main/concordium-cis2/src/lib.rs#L1073">Transfer</a>' used in different smart contracts
+ */
+@Getter
+@AllArgsConstructor
+public class Transfer<T extends TokenId, A extends TokenAmount> {
```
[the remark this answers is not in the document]
magnusbechwind: I used the name `Transfer` to match the struct in https://github.com/Concordium/concordium-rust-smart-contracts/blob/bc7e7ae7db029e78c0f1e965c968ca1364f4ce1c/concordium-cis2/src/lib.rs#L1081C18-L1081C18

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/cis2nft/Cis2NftMethod.java
magnusbechwind: image

concordium-sdk/src/main/java/com/concordium/sdk/crypto/bakertransactions/ConfigureBakerKeysJniInput.java
```diff
@@ -10,10 +10,12 @@
 @Data
 @Builder
 public class ConfigureBakerKeysJniInput {
+
```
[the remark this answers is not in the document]
magnusbechwind: Divine inspiration? No idea haha

concordium-sdk/src/main/java/com/concordium/sdk/requests/smartcontracts/InvokeInstanceRequest.java
```diff
@@ -62,6 +59,101 @@ public class InvokeInstanceRequest {
...
+     * @param amount Amount to invoke the smart contract instance with.
+     * @param schemaParameter {@link SchemaParameter} message to invoke the contract with. Must be initialized with {@link SchemaParameter#initialize()} beforehand.
+     * @param energy The amount of energy to allow for execution. This cannot exceed `100_000_000_000`.
+     */
+    public static InvokeInstanceRequest from(BlockQuery blockHash,
+                                             ContractAddress instance,
+                                             CCDAmount amount,
+                                             SchemaParameter schemaParameter,
+                                             Energy energy) {
+        if (! (schemaParameter.getType() == ParameterType.RECEIVE)) {throw new IllegalArgumentException("SchemaParameter for InvokeInstanceRequest must be initialized with an ReceiveName");}
```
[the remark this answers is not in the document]
magnusbechwind: I think it's the github code view that messed it up because the if statement was on one line - looked fine for me. I reformatted just in case.

concordium-sdk/src/main/java/com/concordium/sdk/requests/smartcontracts/InvokeInstanceRequest.java
```diff
@@ -62,6 +59,101 @@ public class InvokeInstanceRequest {
...
+     * @param schemaParameter {@link SchemaParameter} message to invoke the contract with. Must be initialized with {@link SchemaParameter#initialize()} beforehand.
+     * @param energy The amount of energy to allow for execution. This cannot exceed `100_000_000_000`.
+     */
+    public static InvokeInstanceRequest from(BlockQuery blockHash,
+                                             AbstractAddress invoker,
+                                             ContractAddress instance,
+                                             CCDAmount amount,
+                                             SchemaParameter schemaParameter,
+                                             Energy energy) {
+        if (! (schemaParameter.getType() == ParameterType.RECEIVE)) {throw new IllegalArgumentException("SchemaParameter for InvokeInstanceRequest must be initialized with an ReceiveName");}
```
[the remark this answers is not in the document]
magnusbechwind: So something like "IncokeInstance only supports receive functions"?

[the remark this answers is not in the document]
magnusbechwind: "Cannot initialize smart contract with InvokeInstance. SchemaParameter for InvokeInstanceRequest must be initialized with an ReceiveName" ?

concordium-sdk/src/main/java/com/concordium/sdk/transactions/smartcontracts/parameters/ListParam.java
magnusbechwind: Correct, this is for something like [OperatorOfQueryParams](https://github.com/Concordium/concordium-rust-smart-contracts/blob/bc7e7ae7db029e78c0f1e965c968ca1364f4ce1c/concordium-cis2/src/lib.rs#L1190)
I believe the `#[concordium(transparent]` results in the expected JSON being "just the list" `[OperatorOfQuery, OperatorOfQuery]` and not something like `{queries: [OperatorOfQuery, OperatorOfQuery]}`.  
`List<Foo>` is usable within other parameters without modification e.g. The `List<TokenIdU32>` in `MintParams`

concordium-sdk/src/main/java/com/concordium/sdk/types/UInt8.java
magnusbechwind: It was needed for a smart contract parameters in the examples and we decided to implement it in the sdk as we have UInt16-64.

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/parameters/UpgradeParams.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.io.IOException;
+
+/**
+ * Represents the parameter '<a href="https://github.com/Concordium/concordium-rust-smart-contracts/blob/main/examples/cis2-wccd/src/lib.rs#L150">UpgradeParams</a>' used in the cis2-wCCD contract
+ */
+@Getter
+public class UpgradeParams extends SchemaParameter {
+
+    private final String module;
```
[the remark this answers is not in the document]
magnusbechwind: It is now a `ModuleReference`. To make it work i had to implement custom serialization of `Hash`, but this serialisation does the same as the already present `toString` method, i.e. returns the value as hex, so i believe this will be fine? All tets pass at least.

crypto-jni/src/lib.rs
magnusbechwind: Ive refactored `CryptoJNINative` so the public method for serializing parameters recieves a `SchemaParameter`, `ReceiveName`/`InitName` and a `SchemaVersion`. This ensures the actual rust call is never made with illegal values e.g. 4.

crypto-jni/src/lib.rs
magnusbechwind: Errors are returned as soon as they arise e.g. 
image
image

If the line in question is ever excecuted the serialization was performed correctly

concordium-sdk/src/test/testresources/smartcontractschema/lib.rs
magnusbechwind: As an answer to all of the above. Most of this was generated by `cargo concordium init`. I've documented what is and isn't used

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/cis2nft/Cis2NftMethod.java
EmilLa1: This is OK, but please provide a link to the source. Then users can look up the finer details themselves.

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/parameters/Transfer.java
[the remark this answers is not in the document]
EmilLa1: Yes, but that origins in the CIS2 library. If we at some point want to add another example which has some other `Transfer` then it can end up being a bit confusing. Better to just be explicit in these examples.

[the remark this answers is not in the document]
magnusbechwind: I see, that makes total sense!

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/parameters/UpgradeParams.java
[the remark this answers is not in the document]
EmilLa1: Yes, it should just be the base16 encoded hash.

concordium-sdk/src/test/testresources/smartcontractschema/lib.rs
magnusbechwind: Emil said smart contracts must have errors - it is indeed copy paste from the template.

concordium-sdk/src/test/testresources/smartcontractschema/lib.rs
magnusbechwind: It is never returned from the smart contract. It is just there to make the smart contract valid. Emil said smart contracts must contain errors.

abizjak: You don't have to define a custom type if you don't need it. THere are some predefined errors such as ParseError.

concordium-sdk/src/main/java/com/concordium/sdk/types/UInt8.java
magnusbechwind: Something like this "An unsigned byte, used in the smart contract examples."? Not quite sure i understand what needs to be here.

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/parameters/NftBalanceOfQueryParams.java
```diff
@@ -0,0 +1,16 @@
...
+
+import com.concordium.sdk.transactions.ReceiveName;
+import com.concordium.sdk.transactions.smartcontracts.Schema;
+import com.concordium.sdk.transactions.smartcontracts.parameters.ListParam;
+
+import java.util.List;
+
+/**
+ * Wrapper class for {@link ListParam} enforcing the correct {@link BalanceOfQuery} for cis2-nft contract.
+ */
```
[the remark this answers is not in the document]
magnusbechwind: `ListParam` takes any type of `List`. `NftBalanceOfQueryParams` can only be instantiated with a `List<NftBalanceOfQuery>`

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/parameters/NftTokenMetaDataQueryParams.java
```diff
@@ -0,0 +1,16 @@
...
+import com.concordium.sdk.transactions.smartcontracts.parameters.ListParam;
+
+import java.util.List;
+
+/**
+ * Wrapper class for {@link ListParam} enforcing the correct Token id for cis2-nft contract.
+ */
+public class NftTokenMetaDataQueryParams extends ListParam {
+    public NftTokenMetaDataQueryParams(Schema cis2nftSchema, ReceiveName tokenMetadataReceiveName, List<TokenIdU32> tokensForMetadataQuery) {
+        super(cis2nftSchema, tokenMetadataReceiveName, tokensForMetadataQuery);
```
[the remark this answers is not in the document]
magnusbechwind: Same reasoning as above

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/parameters/OperatorOfQuery.java
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * A query for the operator of a given address for a given token.
+ * Represents a '<a href = "https://github.com/Concordium/concordium-rust-smart-contracts/blob/main/concordium-cis2/src/lib.rs#L1175">OperatorOfQuery</a>' used in different smart contracts
+ */
+@Getter
+@AllArgsConstructor
+public class OperatorOfQuery {
+
+    /**
+     * The ID of the token for which to query the balance of. An {@link AbstractAddress} is either an {@link AccountAddress} or a {@link ContractAddress}.
```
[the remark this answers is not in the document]
magnusbechwind: Shamelessly copy pasted form [concordium-rust-smart-contracts/cis2](https://github.com/Concordium/concordium-rust-smart-contracts/blob/deb6d98e6033edb5e62d9ec445c84bdd3ca323d2/concordium-cis2/src/lib.rs#L1180C29-L1180C29).

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/parameters/SetMetadataUrlParams.java
```diff
@@ -0,0 +1,60 @@
...
+ * Represents the parameter '<a href="https://github.com/Concordium/concordium-rust-smart-contracts/blob/main/examples/cis2-wccd/src/lib.rs#L172">SetMetadataUrlParams</a>' used in the cis2-wCCD contract
+ */
+@Getter
+public class SetMetadataUrlParams extends SchemaParameter {
+
+    private final String url;
+
+    @JsonSerialize(using = SetMetadataUrlParamsHashSerializer.class)
+    private final Optional<Hash> hash;
+    public SetMetadataUrlParams(Schema schema, ReceiveName receiveName, String url, Optional<Hash> hash) {
```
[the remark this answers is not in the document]
magnusbechwind: The constructer of `SchemaPamater` requires a `Schema`, as the serialization is performed by `SchemaParameter`.

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/parameters/TokenAmount.java
```diff
@@ -0,0 +1,7 @@
+package com.concordium.sdk.examples.contractexample.parameters;
+
+/**
+ * Tagging interface for identifying token amounts
+ */
+public interface TokenAmount {
+}
```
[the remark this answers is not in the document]
magnusbechwind: I tried to follow the [smart contract examples](https://github.com/Concordium/concordium-rust-smart-contracts/tree/main/examples) as closesly as possible to hopefully make the translation from "smart contract parameters" to "java smart contract parameters" obvious.

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/parameters/TokenId.java
```diff
@@ -0,0 +1,7 @@
+package com.concordium.sdk.examples.contractexample.parameters;
+
+/**
+ * Tagging interface for identifying tokenIds
+ */
+public interface TokenId {
```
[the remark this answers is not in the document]
magnusbechwind: Same reasoning. Following the rust examples closesly ideally makes it obvious how to implement smart contract parameters using the SDK.

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/parameters/TokenIdU32.java
```diff
@@ -0,0 +1,38 @@
...
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import lombok.Getter;
+import org.apache.commons.codec.binary.Hex;
+
+import java.io.IOException;
+
+@Getter
+@JsonSerialize(using = TokenIdU32.TokenIdU32Serializer.class)
+public class TokenIdU32 implements TokenId {
```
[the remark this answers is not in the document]
magnusbechwind: The inner class `TokenIdU32Serializer` performs the JSON serialization. This is then passed to the rust layer where it is serialized for the contract.

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/parameters/TokenIdU32.java
```diff
@@ -0,0 +1,38 @@
+package com.concordium.sdk.examples.contractexample.parameters;
+
+import com.concordium.sdk.types.UInt32;
```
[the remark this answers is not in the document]
magnusbechwind: Reasoning explained above. Would you prefer just using strings for the example?

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/parameters/WCCDBalanceOfQuery.java
```diff
@@ -0,0 +1,12 @@
...
+
+import com.concordium.sdk.types.AbstractAddress;
+
+/**
+ * Wrapper class for {@link BalanceOfQuery} enforcing the correct Token id for cis2-wCCD contract.
+ */
+public class WCCDBalanceOfQuery extends BalanceOfQuery<TokenIdUnit> {
+    public WCCDBalanceOfQuery(TokenIdUnit tokenId, AbstractAddress address) {
+        super(tokenId, address);
+    }
```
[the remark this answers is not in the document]
magnusbechwind: They specify the typpe of `TokenId`. This is analogous to [ContractBalanceOfQueryParams](https://github.com/Concordium/concordium-rust-smart-contracts/blob/deb6d98e6033edb5e62d9ec445c84bdd3ca323d2/examples/cis2-wccd/src/lib.rs#L842C51-L842C51)

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/parameters/WCCDBalanceOfQueryParams.java
```diff
@@ -0,0 +1,16 @@
...
+
+import java.util.List;
+
+/**
+ * Wrapper class for {@link ListParam} enforcing the correct {@link BalanceOfQuery} for cis2-wccd contract.
+ */
+public class WCCDBalanceOfQueryParams extends ListParam {
+    public WCCDBalanceOfQueryParams(Schema schema, ReceiveName receiveName, List<WCCDBalanceOfQuery> list) {
+        super(schema, receiveName, list);
+    }
```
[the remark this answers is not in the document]
magnusbechwind: This specify the typpe of `BalanceOfQuery`. This is analogous to [ContractBalanceOfQueryParams](https://github.com/Concordium/concordium-rust-smart-contracts/blob/deb6d98e6033edb5e62d9ec445c84bdd3ca323d2/examples/cis2-wccd/src/lib.rs#L842C51-L842C51)

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/parameters/WCCDBalanceOfQuery.java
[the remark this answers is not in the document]
magnusbechwind: This specifies the type of `TokenId` for the `BalanceOfQuery`. You could use `List<BalanceOfQuery<T>>` in `WCCDBalanceOfQueryParams` and avoid this class, but i think instantiating `BalanceOfQuery<TokenIdUnit> = new BalanceOfQuery<TokenIdUnit>(new TokenIdUnit(), ADDRESS)` is very verbose.

concordium-sdk/src/main/java/com/concordium/sdk/ClientV2.java
```diff
@@ -881,6 +883,33 @@ public ImmutableList<WinningBaker> getWinningBakersEpoch(EpochQuery epochQuery)
...
+     * @param transactionHash the {@link Hash} of the transaction to wait for.
+     * @param timeoutMillis the number of milliseconds to wait before throwing an exception.
+     * @return {@link FinalizedBlockItem} of the transaction.
+     * @throws io.grpc.StatusRuntimeException with {@link io.grpc.Status.Code}: <ul>
+     * <li> {@link io.grpc.Status#DEADLINE_EXCEEDED} if the timeout is exceeded.
+     * <li> {@link io.grpc.Status#NOT_FOUND} if the transaction is not known to the node.
+     * </ul>
+     */
+    public FinalizedBlockItem waitUntilFinalized(Hash transactionHash, int timeoutMillis) {
+        BlockItemStatus status = this.getBlockItemStatus(transactionHash);
```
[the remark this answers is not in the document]
magnusbechwind: What do you propose? Should it return an `Optional` instead? Then it would return `Empty` in case the transaction is not found.

concordium-sdk/src/main/java/com/concordium/sdk/transactions/smartcontracts/Schema.java
```diff
@@ -0,0 +1,47 @@
...
+     * The version of the schema.
+     */
+    private final SchemaVersion version;
+
+    /**
+     * Creates a {@link Schema} from the provided bytes and {@link SchemaVersion}.
+     * @param schemaBytes the bytes of the schema.
+     * @param version the version of the schema.
+     */
+    Schema(byte[] schemaBytes, SchemaVersion version) {
```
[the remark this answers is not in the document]
magnusbechwind: Would it make sense to default to V3 if no version is supplied?

concordium-sdk/src/main/java/com/concordium/sdk/transactions/smartcontracts/parameters/AddressParam.java
```diff
@@ -0,0 +1,54 @@
...
+import lombok.Getter;
+
+import java.io.IOException;
+
+/**
+ * Wrapper class allowing {@link AbstractAddress} to be passed directly as a smart contract parameter.
+ */
+@JsonSerialize(using = AddressParam.AddressParamSerializer.class)
+@Getter
+public class AddressParam extends SchemaParameter {
```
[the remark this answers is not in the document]
magnusbechwind: Explained in `AccountAddress` comment

concordium-sdk/src/test/java/com/concordium/sdk/smartcontract/IncorrectParams.java
```diff
@@ -0,0 +1,20 @@
...
+ * Parameters not matching the provided {@link Schema}. Used in {@link SerializeParameterTest} to ensure serialization of incorrect parameters throws an exception.
+ */
+@Getter
+public class IncorrectParams extends SchemaParameter {
+
+    private final int value;
+
+    protected IncorrectParams(Schema schema, ReceiveName receiveName) {
+        super(schema, receiveName);
+        value = 123;
```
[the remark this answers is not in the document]
magnusbechwind: It is just some parameter that does not adhere to the provided `Schema`. the value has no special meaning.

crypto-jni/src/lib.rs
```diff
@@ -408,3 +467,214 @@ pub extern "system" fn Java_com_concordium_sdk_crypto_CryptoJniNative_generateCo
...
+ * Helper method for serialize_init_parameters.
+ * Performs the actual serialization.
+ */
+#[allow(non_snake_case)]
+pub fn serialize_init_parameters_aux(
+    parameter: &str,
+    contractName: &str,
+    schemaBytes: &Vec<u8>,
+    schemaVersion: &Option<u8>,
+    verboseErrors: bool,
```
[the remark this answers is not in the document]
magnusbechwind: What parameter are you refering to? I just see this 
image

crypto-jni/src/lib.rs
```diff
@@ -16,19 +16,22 @@ use concordium_base::transactions::{
...
 use core::slice;
 use ed25519_dalek::*;
-use jni::sys::jstring;
+use jni::sys::{jstring, jboolean, JNI_FALSE};
 use rand::thread_rng;
 use serde_json::{from_str, to_string};
 use std::convert::{From, TryFrom};
 use std::i8;
 use std::str::Utf8Error;
+use hex;
```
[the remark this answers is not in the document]
magnusbechwind: Hex is used to encode the result of serializing the parameter.

crypto-jni/src/lib.rs
```diff
@@ -16,19 +16,22 @@ use concordium_base::transactions::{
...
 
 use jni::{
     objects::{JClass, JString},
     sys::{jbyteArray, jint},
     JNIEnv,
 };
```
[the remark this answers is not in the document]
magnusbechwind: Duplicate suggestion? See above comment.

concordium-sdk/src/main/java/com/concordium/sdk/ClientV2.java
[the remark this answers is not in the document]
abizjak: I think we wait in that case, not fail immediately.

Kind of the same as if it is in "committed" state.

crypto-jni/src/lib.rs
```diff
@@ -408,3 +467,214 @@ pub extern "system" fn Java_com_concordium_sdk_crypto_CryptoJniNative_generateCo
...
+ * Helper method for serialize_receive_parameters.
+ * Performs the actual serialization.
+ */
+pub fn serialize_receive_contract_parameters_aux(
+    parameter: &str,
+    contractName: &str,
+    methodName: &str,
+    schemaBytes: &Vec<u8>,
+    schemaVersion: &Option<u8>,
+    verboseErrors: bool,
```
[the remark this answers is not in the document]
magnusbechwind: It allows for returning more descriptive errors. It is useful for debugging why/if serialization of a parameter fails.

crypto-jni/src/lib.rs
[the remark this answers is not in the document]
magnusbechwind: It allows for returning more descriptive errors. It is useful for debugging why/if serialization of a parameter fails.

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/cis2nft/Cis2NftParameters.java
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    /**
+     * Generates and initializes {@link UpdateOperatorParams} for the 'updateOperator' method of a cis2-nft contract.
+     *
+     * @return initialized {@link UpdateOperatorParams}.
+     */
+    @SneakyThrows
+    public static SchemaParameter generateUpdateOperatorParams() {
+        Schema schema = Schema.from(Files.readAllBytes(Paths.get(SCHEMA_PATH)), SchemaVersion.V3);
```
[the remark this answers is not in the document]
magnusbechwind: The version of the `Schema`. It is needed for the rust layer to serialize the parameters. I see you commented on it further below as well, should it just be V3 as default when initalizing `Schema`s without providing a specific version?.

concordium-sdk/src/test/testresources/smartcontractschema/lib.rs
magnusbechwind: I have removed the custom errors and am now using a standard ParseError

concordium-sdk/src/main/java/com/concordium/sdk/ClientV2.java
```diff
@@ -881,6 +887,79 @@ public ImmutableList<WinningBaker> getWinningBakersEpoch(EpochQuery epochQuery)
...
+     * @param code the {@link io.grpc.Status.Code} to allow.
+     * @throws io.grpc.StatusRuntimeException if the code of the exception does not match the provided code.
+     */
+    private void ignoreExceptionIfCode(io.grpc.StatusRuntimeException e, io.grpc.Status.Code code) {
+        if (e.getStatus().getCode().equals(code)) {
+            return;
+        }
+        throw e;
+    }
+
```
EmilLa1, line 962:
```suggestion
     public Optional<FinalizedBlockItem> waitUntilFinalized(Hash transactionHash, int timeoutMillis) {
        Optional<FinalizedBlockItem> maybeStatus = getFinalizedTransaction(transactionHash);
        // if it's finalized return it.
        if (maybeStatus.isPresent()) {
            return maybeStatus;
        }
        // check newly finalized blocks until we time out
        Optional<FinalizedBlockItem> result = Optional.empty();
        try {
            Iterator<BlockIdentifier> finalizedBlockStream = this.getFinalizedBlocks(timeoutMillis);
            while (finalizedBlockStream.hasNext()) {
                finalizedBlockStream.next();
                Optional<FinalizedBlockItem> finalizedBlockItem = getFinalizedTransaction(transactionHash);
                if (finalizedBlockItem.isPresent()) {
                    // the transaction is included in a finalized block, break and return the finalized status.
                    result = finalizedBlockItem;
                    break;
                }
            }
        } catch (io.grpc.StatusRuntimeException e) {
            // we timed out. Return empty to indicate that the transaction could not be found.
            if (e.getStatus().getCode().equals(Status.Code.DEADLINE_EXCEEDED)) {
                return Optional.empty();
            }
            throw e;
        }
        return result;
    }

    private Optional<FinalizedBlockItem> getFinalizedTransaction(Hash transactionHash) {
        try {
            BlockItemStatus status = this.getBlockItemStatus(transactionHash);
            if (status.getFinalizedBlockItem().isPresent()) {
                return Optional.of(status.getFinalizedBlockItem().get());
            }
            // Only return a finalized transaction.
            return Optional.empty();
        } catch (io.grpc.StatusRuntimeException e) {
            // If the transaction is not found then return empty.
            if (e.getStatus().getCode().equals(io.grpc.Status.Code.NOT_FOUND)) {
                return Optional.empty();
            }
            // report back any other exceptions.
            throw e;
        }
    }

```

I think that this can be written a bit clearer and also I think we should return "empty" in the case where the transaction was not found.

EmilLa1 (commented):
A few extra comments for now

concordium-sdk/src/main/java/com/concordium/sdk/transactions/smartcontracts/parameters/AccountAddressParam.java
```diff
@@ -0,0 +1,41 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import lombok.Getter;
+
+import java.io.IOException;
+
+/**
+ * Wrapper class allowing {@link AccountAddress} to be passed directly as a smart contract parameter.
+ */
+@Getter
+@JsonSerialize(using = AccountAddressParam.AccountAddressParamSerializer.class)
```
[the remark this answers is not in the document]
EmilLa1: This is needed when a parameter is just an 'AccountAddress'. Hence all 'SchemaParameter' types requires the associated 'Schema' when initialized. But a 'AccountAddress' object is very likely not be a 'SchemaParameter' when initialized and hence no accompanying 'Schema' is available. And so this wrapper is required.

But as this type is smart contract dependent (it's an actual parameter to a certain smart contract) then we could also have left out this helper class.

I don't have much knowledge of how frequent this is used as the sole parameter for an init/receive function. If it's very uncommon then I think we should just remove this class (i.e. move it to the actual example where it's needed).

concordium-sdk/src/main/java/com/concordium/sdk/transactions/smartcontracts/parameters/ContractAddressParam.java
```diff
@@ -0,0 +1,41 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import lombok.Getter;
+
+import java.io.IOException;
+
+/**
+ * Wrapper class allowing {@link ContractAddress} to be passed directly as a smart contract parameter.
+ */
+@Getter
+@JsonSerialize(using = ContractAddressParam.ContractAddressParamSerializer.class)
```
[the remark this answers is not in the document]
EmilLa1: Same comment as above. 
1. I think the documentation can be made clearer.
2. Is it required elsewhere than that one example?

concordium-sdk/src/main/java/com/concordium/sdk/transactions/smartcontracts/parameters/AddressParam.java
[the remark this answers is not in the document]
EmilLa1: Same comment as above: 
1. The documentation can be made clearer I think.
2. Is it required by more than just that one example? Maybe it should be moved to the example then in order to avoid confusion.

concordium-sdk/src/main/java/com/concordium/sdk/types/UInt8.java
EmilLa1: So this is a bit better. 
But 1) If it's only used in examples, should it then be here at all? Why not just in the examples then? 
2) This is very specific to a smart contract right? It's not a type that is exposed by the node itself. It could just have exposed an `i8` instead.
3) Is it really needed with this boxed type? Why not just in the smart contract examples use `byte` and then convert to unsigned when doing the serialization?

concordium-sdk/src/test/java/com/concordium/sdk/smartcontract/IncorrectParams.java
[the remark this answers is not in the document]
EmilLa1: Please put explanations in code comments instead of on a comment on github.

concordium-sdk/src/test/testresources/smartcontractschema/lib.rs
```diff
@@ -0,0 +1,142 @@
...
+}
+
+#[derive(Serialize, SchemaType)]
+struct ContractAddressContainer {
+    address:   ContractAddress,
+}
+
+// The `ParseError` in the following result types is never thrown, as the functions only exist to generate a schema for serialization of SC parameters.
+
+/// Takes ListParam as parameter for ensuring correct serialization of ListParam
```
EmilLa1, line 46:
```suggestion
/// Takes [`ListParam`] as parameter for ensuring correct serialization of [`ListParam`]
```
Please use proper rust doc: https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/parameters/TokenIdUnit.java
```diff
@@ -0,0 +1,28 @@
...
+    private final byte id;
+
+    public TokenIdUnit() {
+        id = 0;
+    }
+
+    public static class TokenIdUnitSerializer extends JsonSerializer<TokenIdUnit> {
+        @Override
+        public void serialize(TokenIdUnit value, JsonGenerator gen, SerializerProvider serializers) throws IOException {
+            gen.writeString("");
```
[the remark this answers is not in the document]
magnusbechwind: This is the expected serialization, yes.

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/parameters/TokenIdUnit.java
```diff
@@ -0,0 +1,28 @@
...
+import java.io.IOException;
+
+@Getter
+@JsonSerialize(using = TokenIdUnit.TokenIdUnitSerializer.class)
+public class TokenIdUnit implements TokenId {
+
+    private final byte id;
+
+    public TokenIdUnit() {
+        id = 0;
```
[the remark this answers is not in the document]
magnusbechwind: The serialization was initially the value "0" serialized. But Ales pointed out the empty string is also a valid serialization of this - i just forgot to delete the zero value.

concordium-sdk/src/test/testresources/smartcontractschema/lib.rs
magnusbechwind: As far as i read this is for linking to other rustdcoc? ListParam is refering to the Java class

concordium-sdk/src/main/java/com/concordium/sdk/types/UInt8.java
magnusbechwind: We added it to the SDK for "completeness", as we already have UInt16-32-64. It is only used for these smart contract examples tho, so i believe it would make sense to move it to the examples.

If raw `byte` is used then custom serialization would have to be defined in several places as `UInt8` is used in a lot of different parameters. I believe it is more intuitive, and less error prone, to do it in one place?

concordium-sdk/src/test/testresources/smartcontractschema/lib.rs
EmilLa1: Ahh. maybe clarify that.

concordium-sdk/src/main/java/com/concordium/sdk/types/UInt8.java
magnusbechwind: Removed this class and added `UInt8ByteArrayJsonSerializer` to the examples to handle the serialization of the now used `byte[]`

concordium-sdk/src/main/java/com/concordium/sdk/transactions/smartcontracts/Schema.java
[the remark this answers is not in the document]
magnusbechwind: I've decided to add V3 as the default Schema if no version is supplied.

crypto-jni/src/lib.rs
```diff
@@ -408,3 +465,216 @@ pub extern "system" fn Java_com_concordium_sdk_crypto_CryptoJniNative_generateCo
 
     return CryptoJniResult::Ok(payload).to_jstring(&env);
 }
+
+type SerializeParamResult = CryptoJniResult<String>;
+
+#[no_mangle]
+#[allow(non_snake_case)]
+/// The JNI wrapper for serializing ia parameter for a recieve function according to a specified Schema.
```
abizjak, line 473:
```suggestion
/// The JNI wrapper for serializing ia parameter for a receive function according to a specified Schema.
```

crypto-jni/src/lib.rs
```diff
@@ -408,3 +465,216 @@ pub extern "system" fn Java_com_concordium_sdk_crypto_CryptoJniNative_generateCo
...
+/**
+ * `parameter` The parameter to serialize. `parameter` must be a properly initalized `java.lang.String` that is non-null. The parameter must be valid JSON according to the provided schema.
+ * `contractName` The name of the contract. `contractName` must be a properly initalized `java.lang.String` that is non-null.
+ * `methodName` The name of the method. `methodName` must be a properly initalized `java.lang.String` that is non-null.
+ * `schemaBytes` The bytes of the schema. `schemaBytes` must be a properly initalized `byte[]` containing bytes representing a valid schema. The schema must match the provided `contractName`.
+ * `schemaVersion` The version of the Schema. `schemaVersion` must be an integer representing a valid `com.concordium.sdk.transactions.smartcontracts.SchemaVersion`.
+ * `verboseErrors` Whether errors are returned in verbose format or not, can be useful when debugging why serialization fails.
+ */
```
abizjak, line 481: This is unusual formatting for Rust. You should just use markdown in doc comments

crypto-jni/src/lib.rs
```diff
@@ -408,3 +465,216 @@ pub extern "system" fn Java_com_concordium_sdk_crypto_CryptoJniNative_generateCo
...
+) -> Result<Vec<u8>> {
+    
+    let module_schema = VersionedModuleSchema::new(schemaBytes, &schemaVersion)?;
+    let parameter_type = module_schema.get_init_param_schema(contractName)?;
+    let value: serde_json::Value = from_str(parameter)?;
+
+    let res = parameter_type
+        .serial_value(&value)
+        .map_err(|e| anyhow!("{}", e.display(verboseErrors)));
+    return Ok(res?);
```
abizjak, line 679:
```suggestion
    parameter_type
        .serial_value(&value)
        .map_err(|e| anyhow!("{}", e.display(verboseErrors)))
```

crypto-jni/src/lib.rs
```diff
@@ -408,3 +465,216 @@ pub extern "system" fn Java_com_concordium_sdk_crypto_CryptoJniNative_generateCo
...
+
+    let result = match serializedParameter {
+        Ok(serializedParameter) => hex::encode(serializedParameter),
+        Err(err) => return SerializeParamResult::from(err).to_jstring(&env),
+        
+    };
+
+    return SerializeParamResult::Ok(result).to_jstring(&env);
+
+
```
abizjak, line 656:
```suggestion
```

abizjak (approved):
If it's not there, as it appears not to be, please add clippy & formatting to the rust JNI bindings, using the standard rustfmt.toml config from concordium-base.

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/cis2nft/Cis2NftParameters.java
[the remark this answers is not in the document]
abizjak: I am not a big fan of defaults like this, since they tend to lead to errors. But in this case a schema serialied as V1 or V2 will also parse as a V3 schema (with the same meaning) so it's kind of OK.

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/parameters/SetMetadataUrlParams.java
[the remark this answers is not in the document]
abizjak: Ok, I just thought that in the example these could be specified as constants and only the values can be required as inputs.

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/parameters/TokenAmount.java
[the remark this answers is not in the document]
abizjak: Ok, I don't really think this makes sense in this particular case, but it is not a blocking issue for me.

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/parameters/TokenIdU32.java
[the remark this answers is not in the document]
abizjak: Thank you.

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/parameters/TokenIdU32.java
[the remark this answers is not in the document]
abizjak: No, I would not.

I missed where "above" where the reasoning is.

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/parameters/WCCDBalanceOfQuery.java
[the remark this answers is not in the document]
abizjak: Ok.

concordium-sdk-examples/src/main/java/com/concordium/sdk/examples/contractexample/parameters/SetMetadataUrlParams.java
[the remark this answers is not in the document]
magnusbechwind: A couple of the parameters are used both withing the NFT and the WCCD contract e.g. UpdateOperatorParams

PR #292, opened 2024-01-03
Android sdk

shjortConcordium:
## Purpose

Add support for android / building a AAR artifact.

## Changes

- Added commands to add android targets and build the native lib for those targets in the make file.
- Added new maven module to build AAR artifact for android.
- Added android specific behaviour in NativeResolver.
- Moved package version into a parent module.

## Checklist

      hard-to-understand areas.


.github/workflows/release.yml
```diff
@@ -86,6 +86,30 @@ jobs:
...
+    - name: Setup Rust
+      uses: actions-rs/toolchain@v1
+      with:
+        toolchain: 1.69
+    # Checkout the code
+    - name: Checkout
+      uses: actions/checkout@v2
+      with:
+        submodules: recursive
+    - name: Make android native depencies
```
orhoj, line 102: fix depencies

Makefile
```diff
@@ -23,5 +26,30 @@ all:
 	cp $(PATH_CRYPTO_TARGET)release/crypto_jni.dll $(PATH_JAVA_NATIVE_RESOURCES)
 endif
 
+MIN_ANDROID_VER := "29"
```
orhoj, line 29: Why did you pick 29 here?

README.md
```diff
@@ -68,7 +68,35 @@ One can create a new header file by using the command: `javac -h . ED25519.java`
 
 The output will be a header file, the contents hereof must be matched appropriately as described in the [documentation](https://docs.rs/jni/0.19.0/jni/) into the [lib.rs](crypto-jni/src/lib.rs) rust source file.
 
+## Android
+To build the android AAR package, one has to do the following:
 
+1. Clone this repositry and remember to update submodules, for example when cloning via
```
orhoj, line 74:
```suggestion
1. Clone this repository and remember to update git submodules, for example when cloning via
```

concordium-android-sdk/AndroidManifest.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?>
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+          package="com.concordium.sdk"
+          android:versionCode="1"
+          android:versionName="1.0"
+>
+    <uses-sdk android:minSdkVersion="4" android:targetSdkVersion="16" />
```
orhoj, line 7: The versions here should match what we are building for, right?

concordium-sdk/src/main/java/com/concordium/sdk/crypto/NativeResolver.java
```diff
@@ -19,28 +19,36 @@ public class NativeResolver {
...
-                System.load(tempLib.getAbsolutePath());
                 LOADED = true;
             } catch (Exception e) {
                 throw new IllegalStateException("Could not load native dependencies", e);
             }
         }
     }
 
+    private static boolean isAndroid() {
+        return System.getProperty("java.runtime.name").equals("Android Runtime");
```
orhoj, line 49: Is it documented anywhere that this is the correct way of checking if code is running on Android? I couldn't find anything about it.

An alternative would be to compile a different `NativeResolver` for the Android package vs. the regular one. If the above is not well-documented or guaranteed to be stable.

orhoj (changes requested):

concordium-android-sdk/AndroidManifest.xml
shjortConcordium: I have changed it to 27.

concordium-sdk/src/main/java/com/concordium/sdk/crypto/NativeResolver.java
shjortConcordium: I got it from here :upside_down_face: 
https://stackoverflow.com/questions/4519556/how-to-determine-if-my-app-is-running-on-android
Might not be reliable, but it doesn't seem like there is any "official" way to do it.

I can look into the "compile-time" fix. :thinking:

Makefile
shjortConcordium: I think I copied it from somewhere, I changed it to 27, that's the minimum platform level available on the ubuntu 22.04 image :upside_down_face:

README.md
```diff
@@ -68,7 +68,35 @@ One can create a new header file by using the command: `javac -h . ED25519.java`
 
 The output will be a header file, the contents hereof must be matched appropriately as described in the [documentation](https://docs.rs/jni/0.19.0/jni/) into the [lib.rs](crypto-jni/src/lib.rs) rust source file.
 
+## Android
+To build the android AAR package, one has to do the following:
 
+1. Clone this repository and remember to update git submodules, for example when cloning via
```
EmilLa1, line 74: Are there any prerequisites for building the `.aar` file? E.g. `cargo-ndk` ?

README.md
```diff
@@ -68,7 +68,35 @@ One can create a new header file by using the command: `javac -h . ED25519.java`
...
+2. Set the ANDROID_HOME environment variable to the path to your android SDK installation.
+3. Run `make add-android-targets` from the root of this repository.
+4. Run `make android` from the root of this repository.
+5. Run `mvn install` from the root of the [concordium-android-sdk](./concordium-android-sdk) folder.
+
+`make add-android-targets` adds the rust targets that the native libraries will be built for.
+`make android` builds the native libraries for various targets.
+And the final step builds the aar file `concordium-android-sdk.aar` in the [target](./concordium-android-sdk/target)  folder, which can be used in android projects.
+
+For the library to work on android, the following packages must be added to the project (Shown in the gradle style)
```
EmilLa1, line 87: Would it make sense to build a package where the dependencies are included?

README.md
```diff
@@ -68,7 +68,35 @@ One can create a new header file by using the command: `javac -h . ED25519.java`
 
 The output will be a header file, the contents hereof must be matched appropriately as described in the [documentation](https://docs.rs/jni/0.19.0/jni/) into the [lib.rs](crypto-jni/src/lib.rs) rust source file.
 
+## Android
```
EmilLa1, line 71: Perhaps mention the minimum supported android version.

concordium-sdk/pom.xml
```diff
@@ -4,9 +4,14 @@
...
 
+    <parent>
+        <groupId>com.concordium.sdk</groupId>
+        <artifactId>concordium-sdk-base</artifactId>
+        <version>6.0.1-SNAPSHOT</version>
+    </parent>
+
     <groupId>com.concordium.sdk</groupId>
     <artifactId>concordium-sdk</artifactId>
-    <version>6.0.1-SNAPSHOT</version>
```
EmilLa1, line 9: Does this change anything about the release process? https://github.com/Concordium/concordium-java-sdk/blob/main/release/README.md

E.g. does one now have to run `mvn versions:set -DnewVersion=$VERSION` from the root of the repository? As opposed to the root of the java sdk. ?

How does this affect the release process outlined in `release.yml` ?

EmilLa1 (commented):
Looks good, GJ. 

I have a few comments/questions that I would like to be adressed/answered before merge.

Makefile
orhoj: Note that the current Android wallet has minSdkVersion 26, so this would be more restrictive than what we use there.

.github/workflows/release.yml
```diff
@@ -86,6 +86,30 @@ jobs:
...
+    - name: Build and test android sdk
+      run: cd concordium-android-sdk && mvn --batch-mode --update-snapshots install
+    - name: Upload aar
+      uses: actions/upload-artifact@master
+      with:
+        name: concordium-android-sdk.aar
+        path: ./concordium-android-sdk/target/concordium-android-sdk.aar
+
   build-and-release-jar:
     needs: [build-native-ubuntu, build-native-macos, build-native-windows]
```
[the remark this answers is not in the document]
shjortConcordium: Yes, good catch. :)

README.md
shjortConcordium: Yes :disappointed:
I've added a step to add cargo-ndk, and added more about the Android SDK requirement.

README.md
shjortConcordium: No, As far as I can tell, it seems against best practice, and we should look to deploy it on maven and declare the dependencies there. 
But I just added like this for now. :thinking:

concordium-sdk/pom.xml
shjortConcordium: Yes, it needs to be run from the root. :disappointed:
I've updated the release/README to reflect that.

I've updated the release.yml comments to note that we also get an AAR file in the process, but I don't think it affects it otherwise. :thinking:

Makefile
shjortConcordium: Ahh, that is a good point, I'll test if it works on 26, and set it to that instead.

EmilLa1 (approved):
Looks good! GJ

orhoj (approved):

PR #293, opened 2024-01-05
Key derivation

orhoj:
## Purpose
Add a class `ConcordiumHdWallet` to perform key derivation.

## Changes
- Added `ConcordiumHdWallet` Java class used to derive Concordium keys and randomness.
- Updated Rust JNI library to access key derivation functionality from concordium base.

## Checklist
      hard-to-understand areas.


concordium-sdk/src/main/java/com/concordium/sdk/crypto/wallet/ConcordiumHdWallet.java
```diff
@@ -0,0 +1,328 @@
...
+                maxValue = 255;
+                break;
+            case U32:
+                maxValue = 4294967295l;
+                break;
+            case U64:
+                // Note that a long does not allow for a full u64 as it is a i64.
+                // The types in the SDK would need to be updated to fully accomodate a u64,
+                // and at that point this should also be changed.
+                maxValue = Long.MAX_VALUE;
```
EmilLa1, line 122: Why check for a max value here? If it fits in a long, then it's a 64 bit integer. 

What if the `long` was created with https://docs.oracle.com/javase/8/docs/api/java/lang/Long.html#parseUnsignedLong-java.lang.String- and a value `Long.MAX_VALUE +1`.

EmilLa1 (commented):
Looks good! GJ. 

I have a few comments on the java parts for now, I'll look into the rust parts asap.

crypto-jni/src/lib.rs
```diff
@@ -665,3 +672,445 @@ pub fn serialize_init_parameters_aux(
...
+                errorMessage: err.to_string(),
+                errorType:    JNIErrorResponseType::Utf8Decode,
+            })
+        }
+    };
+
+    Ok(rust_str.to_string())
+}
+
+struct SeedAndNet {
```
EmilLa1, line 700: Please add a bit context to new definitions.

concordium-sdk/src/main/java/com/concordium/sdk/crypto/wallet/ConcordiumHdWallet.java
```diff
@@ -0,0 +1,328 @@
...
+    private String getKeyResult(ExtractKey extractor) {
+        KeyResult result = null;
+        try {
+            String jsonStr = extractor.getKey(this.seedAsHex, this.network.getValue());
+            result = JsonMapper.INSTANCE.readValue(jsonStr, KeyResult.class);
+        } catch (JsonProcessingException e) {
+            throw new RuntimeException(e);
+        }
+
+        if (!result.isSuccess()) {
```
EmilLa1, line 148: Shouldn't this be checked before trying to deserialize the json?

crypto-jni/src/lib.rs
```diff
@@ -665,3 +672,445 @@ pub fn serialize_init_parameters_aux(
...
         .map_err(|e| anyhow!("{}", e.display(verboseErrors)))
 }
+
+fn get_string(env: JNIEnv, java_string: JString) -> Result<String, JNIErrorResponse> {
+    let java_str = match env.get_string(java_string) {
+        Ok(s) => s,
+        Err(err) => {
+            return Err(JNIErrorResponse {
+                errorMessage: err.to_string(),
+                errorType:    JNIErrorResponseType::ParameterSerialization,
```
EmilLa1, line 682: Strictly speaking I believe that this is an instance of a `JsonDeserialization` error.

As the FFI between Java and Rust communicates via Json. 

The `ParameterSerialization` error variant is used for indicating smart contract parameter serialization failure. 

Admittedly, this is not really well documented at the moment.

EmilLa1 (commented):
Looks good. A few minor comments.

concordium-sdk/src/main/java/com/concordium/sdk/crypto/wallet/ConcordiumHdWallet.java
orhoj: The base problem is that there are places in the SDK that uses `long`, when it should be `UInt64`. I don't think that our changes here should be doing that breaking change, so instead I am just trying to protect us from getting bad values.

The check on `maxValue` is redundant when it is a `U64` but it was easier to do the redundant check than to handle it differently. I could refactor it to be handled by itself if you think that is more readable.

If you overflow then the value just becomes negative, and the check `value < 0` should catch that.

concordium-sdk/src/main/java/com/concordium/sdk/crypto/wallet/Network.java
```diff
@@ -0,0 +1,14 @@
+package com.concordium.sdk.crypto.wallet;
+
+public enum Network {
+    Mainnet,
+    Testnet;
```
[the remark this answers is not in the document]
orhoj: It definitely is. I am damage by switching between too many languages. Changing it to all caps.

concordium-sdk/src/main/java/com/concordium/sdk/HexadecimalValidator.java
```diff
@@ -0,0 +1,14 @@
...
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class HexadecimalValidator {
+
+    private static final Pattern HEXADECIMAL_PATTERN = Pattern.compile("\\p{XDigit}+");
+    
+    public static boolean isHexadecimal(String input) {
+        final Matcher matcher = HEXADECIMAL_PATTERN.matcher(input);
```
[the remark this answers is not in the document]
orhoj: As we have the `Hex` class from apache, I have changed the implementation to use that.

concordium-sdk/src/main/java/com/concordium/sdk/crypto/wallet/ConcordiumHdWallet.java
```diff
@@ -0,0 +1,328 @@
...
+        }
+
+        for (long value : values) {
+            if (value > maxValue || value < 0) {
+                throw new IllegalArgumentException("The value " + value + " is not a valid " + type);
+            }
+        }
+    }
+
+    private interface ExtractKey {
```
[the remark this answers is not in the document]
orhoj: I tried to rename and add some documentation.

concordium-sdk/src/main/java/com/concordium/sdk/crypto/wallet/ConcordiumHdWallet.java
orhoj: We had an offline talk and the result of that meant I could remove all this.

crypto-jni/src/lib.rs
orhoj: I have changed the error type.

concordium-sdk/src/main/java/com/concordium/sdk/crypto/wallet/ConcordiumHdWallet.java
orhoj: Without deserializing I would not be able to inspect the field. What I have done here is basically a copy-paste of the approach used elsewhere in the SDK. You can search for `isSuccess()` to see how it is elsewhere as well.

concordium-sdk/src/main/java/com/concordium/sdk/crypto/wallet/WordLists.java
```diff
@@ -0,0 +1,2058 @@
+package com.concordium.sdk.crypto.wallet;
+
+public class WordLists {
+    /*
+     * The English BIP39 wordlist as taken from
+     * https://github.com/bitcoin/bips/blob/master/bip-0039/english.txt.
+     * This wordlist is in code as it is supplied as a resource file in the BIP39
```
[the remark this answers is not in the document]
orhoj: It could be added as a resource but by having the list in code means we don't have to do any special Android handling. At this time I think I slightly prefer just having it in-code, but if you think it is nicer to add as a resource for Android, then I can take a look at that.

EmilLa1 (approved):
Looks good!

PR #298, opened 2024-01-22
Update concordium-base dependency, rand and ed25519 dependencies

orhoj:
## Purpose
Concordium-base was updated (as we need some new functionality from wallet_library), and as concordium-base has updated some dependencies we also use, they had to be updated to be in sync.

## Changes
- Updated dependency to `concordium-base`.
- Updated `rand`, `ed25519-dalek` and `curve25519-dalek` to the versions also used by `concordium-base`.

## Checklist

      hard-to-understand areas.


crypto-jni/Cargo.toml
```diff
@@ -7,9 +7,9 @@ edition = "2018"
 
 [dependencies]
 jni = "0.19.0"
-rand = "=0.7"
-ed25519-dalek = "=1.0"
-curve25519-dalek = "3.0"
+rand = "=0.8"
+ed25519-dalek = "=2.0"
+curve25519-dalek = "4.1"
```
abizjak, line 12: I'm surprised the curve is needed as a direct dependency. I can't seem to find where it is used.

crypto-jni/src/lib.rs
```diff
@@ -103,7 +100,13 @@ pub extern "system" fn Java_com_concordium_sdk_crypto_CryptoJniNative_verify(
         Ok(x) => x,
         _ => return NATIVE_CONVERSION_ERROR,
     };
-    let public_key = match PublicKey::from_bytes(&public_key_bytes) {
+
+    let public_key_bytes_array: [u8; 32] = match public_key_bytes.try_into() {
+        Ok(s) => s,
+        Err(_) => return MALFORMED_SECRET_KEY,
```
abizjak, line 106: This does not seem the right error.

abizjak (approved):

crypto-jni/Cargo.toml
orhoj: I just updated it as it was already there. I have just removed it now, so you are right in that it was not in use.

crypto-jni/src/lib.rs
orhoj: Whoops. Changed to `MALFORMED_PUBLIC_KEY`.

Issue #306, opened 2024-02-05
Error in `getAccountInfo` return the `IllegalArgumentException` for any address with delegation and without active pending changes


horus20:
**Bug Description**
When I request information about any addresses with active delegation, but without pending operation I receive the exception.
In the `ClientV2MapperExtensions` interface I found 3 similar places 
```.pendingChange(to(stake.getPendingChange()))```
and if we don't have pending stake operations we get here default instance, of course it's not correct.
In one place this moment already fixed
```
.pendingChange(delegatorInfo.hasPendingChange()
    ? Optional.of(to(delegatorInfo.getPendingChange()))
    : Optional.empty())
```
I guess, this is more correct, and need to copy this approach to other similar places.

**Steps to Reproduce**
Request the `getAccountInfo` from api v2 for some address with the delegation, for example: 4pcKwXGJFWrWMMnmoGrNk4upY9j8vBEFgXVNzgJvKdEANTHW2i (testnet)

**Expected Result**
Account information with the information about delegation.

**Actual Result**
`IllegalArgumentException` in this method `ClientV2MapperExtensions` method `static PendingChange to(StakePendingChange pendingChange)` 

**Versions**
- SDK Version: 6.0.0 and 6.1.0

EmilLa1:
Hi @horus20 
Thanks for reporting this issue. This is indeed a bug as there might not be any pending changes as you correctly point out.

PR #307, opened 2024-02-05
Fix bug where AccountInfo could fail if there were no pendingchange for a delegator or validator.

EmilLa1:
## Purpose

Fix https://github.com/Concordium/concordium-java-sdk/issues/306

## Changes

- Fixed  parsing of the delegator / baker stake objects
- `Baker` and `AccountDelegation` now exposes `Optional<PendingChange>` as opposed to just
  `PendingChange`.
- Fix `.equals` on `AccountInfo`
- Added a test

## Checklist

- [ ] My code follows the style of this project.
- [ ] The code compiles without warnings.
- [ ] I have performed a self-review of the changes.
- [ ] I have documented my code, in particular the intent of the
      hard-to-understand areas.
- [ ] (If necessary) I have updated the CHANGELOG.

### DIFF
--- a/concordium-sdk/src/main/java/com/concordium/sdk/ClientV2MapperExtensions.java
+++ b/concordium-sdk/src/main/java/com/concordium/sdk/ClientV2MapperExtensions.java
@@ -27,7 +27,6 @@ import com.concordium.sdk.crypto.ed25519.ED25519PublicKey;
 import com.concordium.sdk.crypto.elgamal.ElgamalPublicKey;
 import com.concordium.sdk.crypto.pedersencommitment.PedersenCommitmentKey;
 import com.concordium.sdk.crypto.pointchevalsanders.PSPublicKey;
-import com.concordium.sdk.crypto.wallet.credential.CredentialDeploymentDetails;
 import com.concordium.sdk.requests.AccountQuery;
 import com.concordium.sdk.requests.BlockQuery;
 import com.concordium.sdk.requests.EpochQuery;
@@ -325,7 +324,7 @@ interface ClientV2MapperExtensions {
     @Nullable
     static AccountDelegation to(com.concordium.grpc.v2.AccountStakingInfo.Delegator stake) {
         return AccountDelegation.builder()
-                .pendingChange(to(stake.getPendingChange()))
+                .pendingChange(stake.hasPendingChange() ? Optional.of(to(stake.getPendingChange())) : Optional.empty())
                 .restakeEarnings(stake.getRestakeEarnings())
                 .stakedAmount(to(stake.getStakedAmount()))
                 .target(to(stake.getTarget()))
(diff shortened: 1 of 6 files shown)


concordium-sdk/src/main/java/com/concordium/sdk/ClientV2MapperExtensions.java
```diff
@@ -874,13 +873,13 @@ static SendBlockItemRequest to(CredentialDeploymentTransaction credentialDeploym
...
-                    .setRawPayload(ByteString.copyFrom(credentialDeploymentTransaction.getPayloadBytes()))
-                    .build()
+                .setCredentialDeployment(
+                        CredentialDeployment.newBuilder()
+                                .setMessageExpiry(time)
+                                .setRawPayload(ByteString.copyFrom(credentialDeploymentTransaction.getPayloadBytes()))
+                                .build()
                 )
-            .build();
+                .build();
```
abizjak, line 882: This does not look like a relevant change.

concordium-sdk/src/test/java/com/concordium/sdk/ClientV2GetAccountInfoTest.java
```diff
@@ -113,176 +116,193 @@ private static Commitment toCommitment(byte[] commitmentCredCounter) {
...
+                                commissionRates(com.concordium.sdk.responses.accountinfo.CommissionRates.builder().
+                                        bakingCommission((double) COMMISSION_BAKING_PPHT / 100_000D).
+                                        finalizationCommission((double) COMMISSION_FINALIZATION_PPHT / 100_000D).
+                                        transactionCommission((double) COMMISSION_TRANSACTION_PPHT / 100_000D).
+                                        build()).
+                                build()).
+                        build()).
+                build();
+
+    }
```
abizjak, line 305: Why is all of this reformatted? What even is the change?

concordium-sdk/src/main/java/com/concordium/sdk/ClientV2MapperExtensions.java
```diff
@@ -369,7 +368,7 @@ static PendingChange to(StakePendingChange pendingChange) {
                         .build();
             default:
             case CHANGE_NOT_SET:
```
abizjak, line 370: In what cases is this case reached? Who has to fail?

abizjak (approved):

concordium-sdk/src/main/java/com/concordium/sdk/ClientV2MapperExtensions.java
EmilLa1: Not sure why it's been reformatted. Changed it back.. for now..

concordium-sdk/src/main/java/com/concordium/sdk/ClientV2MapperExtensions.java
EmilLa1: As per my understanding, then it is reached if the pending change is not set. 
https://protobuf.dev/reference/java/java-generated/#oneof-fields

But before calling this function we always make sure that it is indeed set.

However, cleaned this up now, so this function returns `Optional<..>` and the `Optional.empty()` is returned via the case mentioned.

PR #313, opened 2024-02-12
Check for redirect before opening browser

orhoj:
## Purpose
Align the Android example with the flow from the web example

## Changes
- Make it so that the Android example wallet opens the browser at the redirected URL instead of letting the browser handle the redirect.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-android-wallet-example/app/src/main/java/com/concordium/example/wallet/activities/IssueIdentityActivity.kt
+++ b/concordium-android-wallet-example/app/src/main/java/com/concordium/example/wallet/activities/IssueIdentityActivity.kt
@@ -30,12 +30,20 @@ import com.concordium.example.wallet.Requests
 import com.concordium.example.wallet.Storage
 import com.concordium.example.wallet.services.ConcordiumClientService
 import com.concordium.example.wallet.services.ConcordiumWalletProxyService
+import com.concordium.example.wallet.services.IdentityFetcherService
 import com.concordium.example.wallet.services.wallet_proxy.IdentityProvider
 import com.concordium.example.wallet.ui.Container
 import com.concordium.example.wallet.ui.Menu
+import com.google.api.Http
+import com.google.common.net.HttpHeaders
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+import okhttp3.OkHttpClient
+import okhttp3.Request
+import retrofit2.http.HTTP
 
 class IssueIdentityActivity : ComponentActivity() {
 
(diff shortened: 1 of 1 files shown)


concordium-android-wallet-example/app/src/main/java/com/concordium/example/wallet/activities/IssueIdentityActivity.kt
```diff
@@ -68,10 +76,30 @@ class IssueIdentityActivity : ComponentActivity() {
...
+                val url = getIssuanceUrl(provider, request)
+
+                val okHttpClientBuilder = OkHttpClient().newBuilder()
+                    .followRedirects(false)
+                    .followSslRedirects(false)
+                val client = okHttpClientBuilder.build()
+                val httpRequest = Request.Builder().url(url).build()
+
+                client.newCall(httpRequest).execute().use { response ->
+                    val redirectedUrl = response.header(HttpHeaders.LOCATION)
```
Radiokot, line 91: I'm trying to do the same for CryptoX, and returning the header turned out to be not enough. `Location` may have either absolute or relative URL. The second case won't work. Can be tested with Notabene staging.

Radiokot: Found out that in order to get the latest redirect URL, you can let the HTTP client follow redirects and query `Response.networkResponse`:
```kotlin
    fun getVerificationRedirectUrl(verificationStartUrl: String): String =
        httpClient
            .newCall(
                Request.Builder()
                    .url(verificationStartUrl)
                    .cacheControl(CacheControl.FORCE_NETWORK)
                    .build()
            )
            .execute()
            .networkResponse
            ?.request
            ?.url // This is the last requested URL after following all the redirects.
            ?.toString()
            ?: error("The identity provider did not redirect as expected")
```

orhoj: I will take a look at this tomorrow, but it was intentional to not follow redirects as there should be exactly one.

Radiokot: My updated solution:
```kotlin
fun getVerificationRedirectUri(
    verificationStartUrl: String,
    redirectUriScheme: String,
): String =
    httpClient
        .newBuilder()
        .followRedirects(false)
        .followSslRedirects(false)
        .build()
        .newCall(
            Request.Builder()
                .url(verificationStartUrl)
                .cacheControl(CacheControl.FORCE_NETWORK)
                .build()
        )
        .execute()
        .let { response ->
            check(response.isRedirect) {
                "The identity provider did not redirect as expected"
            }

            val redirectLocation = response.header("Location")
                ?: error("Can't find the location in the identity provider response")

            return@let if (Uri.parse(redirectLocation).scheme in setOf(
                    "http",
                    "https",
                    redirectUriScheme,
                )
            ) {
                // Return the location if it is a URI with an allowed scheme:
                // HTTP or HTTPS to proceed with the verification on a web page,
                // or the app redirect URI scheme to get the instant result.
                redirectLocation
            } else {
                // Otherwise try to resolve the location against the base URL
                // in case it is a relative path.
                response.request.url.resolve(redirectLocation)?.toString()
                    ?: error("Can't resolve the redirect: $redirectLocation")
            }
        }
```

Radiokot: By the way, found out that Digital Trust shoots `409 Conflict` instead of redirect in case of idCredPub duplication 🙃

orhoj: I will make a writeup of the issues with the identity providers, and then just pause this issue for now. If other wallets are going to implement this, then we need the protocol to be well-defined.

PR #321, opened 2024-02-20
Qualified statement

shjortConcordium:
## Purpose

Add type for the statements that is received from dApps (unqualified) and add methods to convert them into the ones that are needed for proofs (qualified).

### DIFF
--- a/concordium-sdk/src/main/java/com/concordium/sdk/crypto/wallet/web3Id/Statement/UnqualifiedRequestStatement.java
+++ b/concordium-sdk/src/main/java/com/concordium/sdk/crypto/wallet/web3Id/Statement/UnqualifiedRequestStatement.java
@@ -27,7 +27,7 @@ public class UnqualifiedRequestStatement extends RequestStatement {
     
     public QualifiedRequestStatement qualify(CredentialRegistrationId credId, Network network) {
         if (!idQualifier.getType().equals(StatementType.Credential)) {
-            throw new IllegalArgumentException("Only an account statement may only be qualified using a credentialId");
+            throw new IllegalArgumentException("Only an account statement may be qualified using a credentialId");
         }
         RequestIdentifier did = new AccountRequestIdentifier(network, credId);
         return QualifiedRequestStatement.builder().id(did).statement(getStatement()).build();
@@ -36,7 +36,7 @@ public class UnqualifiedRequestStatement extends RequestStatement {
     public QualifiedRequestStatement qualify(ContractAddress contract, ED25519PublicKey publicKey, Network network) {
         if (!idQualifier.getType().equals(StatementType.SmartContract)) {
             throw new IllegalArgumentException(
-                    "Only an web3Id statement may only be qualified using a Contract Address");
+                    "Only a verifiable credential statement may be qualified using a Contract Address");
         }
         if (!((VerifiableCredentialQualifier) idQualifier).getIssuers().contains(contract)) {
             throw new IllegalArgumentException("The Contract Address must be one specified in the qualifier");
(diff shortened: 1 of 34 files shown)


concordium-sdk/src/main/java/com/concordium/sdk/crypto/wallet/web3Id/Statement/Web3IdIssuerQualifier.java
orhoj, line 1: I suggest that we name this `VerifiableCredentialQualifier` to align with the web-sdk.

concordium-sdk/src/main/java/com/concordium/sdk/crypto/wallet/web3Id/Statement/UnqualifiedRequestStatement.java
```diff
@@ -0,0 +1,47 @@
...
+
+@Getter
+@SuperBuilder
+@Jacksonized
+public class UnqualifiedRequestStatement extends RequestStatement {
+    private IdQualifier idQualifier;
+
+    public QualifiedRequestStatement qualify(CredentialRegistrationId credId, Network network) {
+        if (!idQualifier.getType().equals(IdQualifier.QualifierType.Credential)) {
+            throw new IllegalArgumentException("Only an account statement may only be qualified using a credentialId");
```
orhoj, line 22:
```suggestion
            throw new IllegalArgumentException("Only an account statement may be qualified using a credentialId");
```

concordium-sdk/src/main/java/com/concordium/sdk/crypto/wallet/web3Id/Statement/UnqualifiedRequestStatement.java
```diff
@@ -0,0 +1,47 @@
...
+        }
+        String did = "did:ccd:" + network.getValue().toLowerCase() + ":cred:" + credId.getEncoded();
+        return QualifiedRequestStatement.builder().id(did).statement(getStatement()).build();
+    }
+
+    public QualifiedRequestStatement qualify(ContractAddress contract, ED25519PublicKey publicKey, Network network) {
+        if (!idQualifier.getType().equals(IdQualifier.QualifierType.SmartContract)) {
+            throw new IllegalArgumentException(
+                    "Only an web3Id statement may only be qualified using a Contract Address");
```
orhoj, line 31:
```suggestion
                    "Only a verifiable credential statement may be qualified using a Contract Address");
```

concordium-sdk/src/main/java/com/concordium/sdk/crypto/wallet/web3Id/Statement/UnqualifiedRequestStatement.java
```diff
@@ -0,0 +1,47 @@
...
+        }
+        if (!((Web3IdIssuerQualifier) idQualifier).getIssuers().contains(contract)) {
```
orhoj, line 33: If we check this here, then perhaps we should also do the equivalent check when qualifying an account credential statement.

orhoj (changes requested):

concordium-sdk/src/main/java/com/concordium/sdk/crypto/wallet/web3Id/UnqualifiedRequest.java
```diff
@@ -0,0 +1,33 @@
...
+     * @return a QualifiedRequest with the same challenge, and with the statements
+     *         transformed using the qualifier.
+     */
+    public QualifiedRequest qualify(Qualifier qualifier) {
+        List<QualifiedRequestStatement> qualifiedStatements = credentialStatements.stream().map(qualifier::qualify)
+                .collect(Collectors.toList());
+        return QualifiedRequest.builder().challenge(challenge).credentialStatements(qualifiedStatements).build();
+    }
+
+    static interface Qualifier {
```
orhoj, line 30:
```suggestion
    public static interface Qualifier {
```

concordium-sdk/src/main/java/com/concordium/sdk/crypto/wallet/web3Id/Statement/UnqualifiedRequestStatement.java
shjortConcordium: Yes, but then we would just need to add more info, I just added it here because we can check it with only the info needed to qualify the statement.

shjortConcordium:
I'll address the comments/fix stuff when we finish with the `can-prove-statement`.

orhoj (approved):

PR #328, opened 2024-03-13
Fetch schema

magnusbechwind:
## Purpose

Add convenience method to exctract a `Schema` from a `WasmModule` if embedded.

## Changes

I added the class `LEB128U` to encode/decode LEB128U encoded integers and refactored `TokenAmount` to use this.

I added the method `getSchema` to `WasmModule` that parses the underlying wasm bytes and exctract a `Schema` if found.

I added a test extracting a `Schema` and asserting it works.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/concordium-sdk/src/main/java/com/concordium/sdk/transactions/smartcontracts/WasmModule.java
+++ b/concordium-sdk/src/main/java/com/concordium/sdk/transactions/smartcontracts/WasmModule.java
@@ -127,8 +127,8 @@ public class WasmModule {
      * @return {@link Optional} containing the {@link Schema} if found, empty otherwise.
      */
     public Optional<Schema> getSchema() {
-        val moduleSourceBytes = source.getBytes();
-        val buffer = ByteBuffer.wrap(moduleSourceBytes.clone());
+        val moduleSourceBytes = source.getBytes().clone();
+        val buffer = ByteBuffer.wrap(moduleSourceBytes);
 
         // Skip 4 byte length of WasmModuleSource (UInt32.BYTES) + 4 byte magic number + 4 byte WASM version
         buffer.position(buffer.position() + 12);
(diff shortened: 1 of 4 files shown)


concordium-sdk/src/main/java/com/concordium/sdk/transactions/smartcontracts/WasmModule.java
```diff
@@ -83,6 +86,17 @@ public static WasmModule from(final byte[] bytes) {
         return from(moduleBytes, version);
     }
 
+    /**
+     * Create {@link WasmModule} from compiled module WASM file. Passed module should have version prefixed.
```
abizjak, line 90: Clarify that this is the module format produced by cargo-concordium.

concordium-sdk/src/main/java/com/concordium/sdk/transactions/smartcontracts/WasmModule.java
```diff
@@ -107,4 +121,55 @@ public byte[] getBytes() {
...
+
+            byte[] nameBytes = new byte[nameLength];
+            buffer.get(nameBytes);
+
+            String name = Strings.fromByteArray(nameBytes);
+
+            // We've incremented the buffer by reading the length of the name and the name.
+            remainingSectionLength = remainingSectionLength - nameLengthBytes - nameLength;
+
+            if (name.equals("concordium-schema")) {
```
abizjak, line 160: In principle there are also `concordium-schema-v1` and `concordium-schema-v2` , see https://docs.rs/axum/latest/axum/struct.Router.html#method.fallback_service https://github.com/Concordium/concordium-base/blob/87a752f185869ea567387baac384e0e5bc997fd9/smart-contracts/wasm-chain-integration/src/utils.rs#L583

which would be great to support since the Schema class does as well as far as I understand.

The difference is that the `concordium-schema` section contains the schema that is versioned inside.

Whereas the new (old,legacy) `-v1` and `-v2` sections, the contents is the unversioned schema and the version is in the name (v0 and v1).

abizjak (commented):
Looks pretty good. But I think we should add support for legacy schemas as well for completeness.

PR #341, opened 2024-08-28
JAV-319 Support for Protocol 7

Radiokot:
## Purpose

Support Protocol 7.

## Changes

- Added `ProtocolVersion.V7` corresponding to Protocol version 7
- Added `cooldowns` list to `AccountInfo`
- Added `availableBalance` to `AccountInfo`
- Made optional the following `BakerPoolStatus` fields: 
  `bakerEquityCapital`, `delegatedCapital`, `delegatedCapitalCap`, `poolInfo`
- Updated version to `7.3.0-SNAPSHOT`

## Checklist

      hard-to-understand areas.


concordium-android-sdk/pom.xml
```diff
@@ -5,7 +5,7 @@
     <parent>
         <groupId>com.concordium.sdk</groupId>
         <artifactId>concordium-sdk-base</artifactId>
-        <version>7.2.0</version>
+        <version>7.3.0-SNAPSHOT</version>
```
soerenbf, line 8: Surely these will be breaking changes, right? Maybe you should 
```suggestion
        <version>8.0.0-SNAPSHOT</version>
```

concordium-sdk/src/main/java/com/concordium/sdk/responses/poolstatus/BakerPoolStatus.java
```diff
@@ -60,6 +61,22 @@ public class BakerPoolStatus {
...
+    }
+
+    public Optional<CCDAmount> getDelegatedCapitalCap() {
+        return Optional.ofNullable(delegatedCapitalCap);
+    }
+
+    public Optional<BakerPoolInfo> getPoolInfo() {
+        return Optional.ofNullable(poolInfo);
+    }
+
```
soerenbf, line 79: I don't know a lot of java, but how would you even access these before? Is there something implicit getter implementation going on with one of the annotations `@Data` for example? and does the implementation of `public Optional<X> getY()` override the implementation otherwise implicitly provided?

soerenbf (commented):
IMO looks good. You need to handle the newly added transaction events however.

Have a look at the added transaction events here: [BakerEvent.DelegationRemoved](https://github.com/Concordium/concordium-grpc-api/compare/grpc-api/6.2...grpc-api/7.0#diff-a7337993ee30524fd2e3ee681d92ca3a483bb59389f312a741822917033175fdR1199) and [DelegationEvent.BakerRemoved](https://github.com/Concordium/concordium-grpc-api/compare/grpc-api/6.2...grpc-api/7.0#diff-a7337993ee30524fd2e3ee681d92ca3a483bb59389f312a741822917033175fdR1262)

Radiokot:
It seems that the SDK doesn't handle any events, they are only found in gRPC generated classes.

concordium-android-sdk/pom.xml
Radiokot: No breaking changes so far.

concordium-sdk/src/main/java/com/concordium/sdk/responses/poolstatus/BakerPoolStatus.java
Radiokot: There is a plugin which generates getters, but for fields which sometimes are expected to be `null` there are explicit getters returning `Optional`, they are added manually. It is not my idea, I followed the code written last year by Emil Lai and Prav Sharma.

soerenbf: Yeah I read a little about `@Data` and `@Value` annotations from `lombok` yesterday 😄

Changing a function signature from `CCDAmount getBakerEquityCapital()` to `Optional<CCDAmount> getBakerEquityCapital()` _is_ a breaking change in my mind though.. can you explain why you think it is not?

soerenbf:
> It seems that the SDK doesn't handle any events, they are only found in gRPC generated classes.

I don't understand why I cannot find the added types anywhere then? To me it looks like anything in https://github.com/Concordium/concordium-java-sdk/tree/JAV-319-protocol-7/concordium-sdk/src/main/java/com/concordium/sdk/responses contains classes which are created from the generated GRPC types, and this folder https://github.com/Concordium/concordium-java-sdk/tree/JAV-319-protocol-7/concordium-sdk/src/main/java/com/concordium/sdk/responses/transactionstatus is supposed to (I assume) have 2 extra classes for the respective events linked in my original review:

> Have a look at the added transaction events here: [BakerEvent.DelegationRemoved](https://github.com/Concordium/concordium-grpc-api/compare/grpc-api/6.2...grpc-api/7.0#diff-a7337993ee30524fd2e3ee681d92ca3a483bb59389f312a741822917033175fdR1199) and [DelegationEvent.BakerRemoved](https://github.com/Concordium/concordium-grpc-api/compare/grpc-api/6.2...grpc-api/7.0#diff-a7337993ee30524fd2e3ee681d92ca3a483bb59389f312a741822917033175fdR1262)

And of course these would have to be mapped as well somewhere.

concordium-sdk/src/main/java/com/concordium/sdk/responses/poolstatus/BakerPoolStatus.java
Radiokot: Of course, you are right! I explained how the getters work but haven't seen a breaking change introduced by them 😅

concordium-android-sdk/pom.xml
Radiokot: Actually, there are breaking changes in `BakerPoolStatus` getters, so I'll change the version to 8.0.0-SNAPSHOT

Radiokot:
> I don't understand why I cannot find the added types anywhere then? I guess the generated code is not part of the repository?

Yes, they get generated locally:
image

Classes under the `responses` package are manually written human-friendly mappings of GRPC classes involved in `ClientV2` calls. The conversion between GRPC and responses is done in `ClientV2MapperExtensions`.

I'll take a look at `transactionstatus`.

concordium-sdk/src/main/java/com/concordium/sdk/responses/transactionstatus/BakerDelegationRemoved.java
```diff
@@ -0,0 +1,34 @@
...
+public class BakerDelegationRemoved extends AbstractBakerResult {
+
+    private final AccountIndex delegatorId;
+
+    public static BakerDelegationRemoved from(BakerEvent.DelegationRemoved bakerDelegationRemoved, AccountAddress sender) {
+        return BakerDelegationRemoved
+                .builder()
+                .delegatorId(AccountIndex.from(bakerDelegationRemoved.getDelegatorId()))
+                .account(sender)
+                .build();
```
soerenbf, line 27: In the original event, there is not `AccountAddress` in the data. I think we should keep to that here as well 🙂

concordium-sdk/src/main/java/com/concordium/sdk/responses/transactionstatus/DelegationBakerRemoved.java
```diff
@@ -0,0 +1,30 @@
...
+
+    private final long bakerId;
+
+    public static DelegationBakerRemoved from(DelegationEvent.BakerRemoved delegationBakerRemoved, AccountAddress sender) {
+        return DelegationBakerRemoved
+                .builder()
+                .bakerId(delegationBakerRemoved.getBakerId().getValue())
+                .delegatorAddress(sender)
+                .build();
+    }
```
soerenbf, line 24: Same here

soerenbf (approved):
Some minor inconsistencies. When they're fixed, feel free to merge 😄

concordium-sdk/src/main/java/com/concordium/sdk/responses/transactionstatus/BakerDelegationRemoved.java
Radiokot: For all the already existing events, this address is set to the transaction `sender` address and not taken from the event data. For example, none of `BakerEvent`s have `account` attribute https://github.com/Concordium/concordium-grpc-api/blob/88183a8f5e0a3208f4da65d46e0c56d7ec5f14f7/v2/concordium/types.proto#L1133

Maybe the reason for adding `account` is that this model represents not an event but a transaction result, therefore it makes sense to know who was the sender. I wish we could ask the creators.

soerenbf: You're right - let's just continue as is then 😄

PR #350, opened 2025-01-21
Protocol 8

Radiokot:
## Purpose

Support Protocol 8 and the latest GRPC changes.

## Changes

- Added `ProtocolVersion.V8` corresponding to Protocol version 8
- Added `BakerSuspended` and `BakerResumed` transaction result events
- Added `ValidatorSuspended` and `ValidatorPrimedForSuspension` special outcomes
  for the `GetBlockSpecialEvents` endpoint
- Added `ChainParametersV3` for Protocol version 8, adding `ValidatorScoreParameters`
- Added `GetConsensusDetailedStatus` endpoint for querying detailed consensus status information
- Added `GetScheduledReleaseAccounts` endpoint for querying the list of accounts that have scheduled releases
- Added `GetCooldownAccounts`, `GetPreCooldownAccounts` and `GetPrePreCooldownAccounts`
  endpoints for querying the lists of accounts that have pending cooldowns in protocol version 7 onwards
- Added `parameter` field to `ContractInitializedResult` containing the parameter passed to the init function
- Added `isSuspended` field to `Baker` and `BakerPoolStatus` which, since Protocol version 8,
  indicates whether the account is suspended and is not participating in the consensus algorithm
- Added `missedRounds` and `isPrimedForSuspension` fields to `CurrentPaydayStatus` 
  which, since Protocol version 8, reflect validator score status
- Added `suspended` field to `ConfigureBakerPayload` which, since Protocol version 8,
  indicates whether the validator is suspended
- Added `FinalizationCommitteeParameters` and `ValidatorScoreParameters` entries of `PendingUpdateType`, 
  therefore new possible update types in `PendingUpdateV2`
- Added `validatorScoreParameters` field to `NextUpdateSequenceNumbers`
  which corresponds to updates in `ChainParametersV3.validatorScoreParameters`

## Checklist

      hard-to-understand areas.


concordium-sdk/src/main/java/com/concordium/sdk/responses/transactionstatus/ContractInitializedResult.java
```diff
@@ -65,11 +71,12 @@ public static ContractInitializedResult from(ContractInitializedEvent contractIn
                 .amount(CCDAmount.from(contractInitialized.getAmount()))
                 .version(ContractVersion.from(contractInitialized.getContractVersion()))
                 .initName(contractInitialized.getInitName().getValue())
+                .parameter(Parameter.from(contractInitialized.getParameter()))
```
soerenbf, line 74: This has nothing to do with p8 I assume?

concordium-android-sdk/pom.xml
```diff
@@ -5,7 +5,7 @@
     <parent>
         <groupId>com.concordium.sdk</groupId>
         <artifactId>concordium-sdk-base</artifactId>
-        <version>8.0.0</version>
+        <version>8.1.0</version>
```
soerenbf, line 8: Is adding more enum variants not a breaking change? I would expect version `9.0.0` 🤔

soerenbf (commented):
You weren't kidding when you said there was a lot of ground to cover... 😅 

I feel like much of it stems from the addition of the detailed consensus status query which was not implemented for the other SDK's but only for `concordium-client`. Great job though, also with tests and documentation.

A few things I could not find:
- new `PendingUpdate` variant (is this query not provided by the client?)
- new `NextUpdateSequenceNumber` field (I guess we don't support updates in this SDK?)
- new parameter update transaction for updating the validator score parameter (I guess the same reason as above)

concordium-sdk/src/main/java/com/concordium/sdk/responses/transactionstatus/ContractInitializedResult.java
Radiokot: Yep, this is a part of GRPC changelog, back then I assumed everything new in GRPC is related to Protocol 8

concordium-android-sdk/pom.xml
Radiokot: Yeah, you're right, we can't be sure that all the developers who use Java SDK have implemented graceful failing when encountering unknown enum entries. Although it will compile in Java, the runtime behavior could be unexpected.

Radiokot:
Thanks, Soren, I addressed all your comments except this one:

> new parameter update transaction for updating the validator score parameter

You're right, the chain params update transaction is not provided by the SDK.

soerenbf (approved):
LGTM!

